PR #59, opened 2015-03-04
Add path prefix

miaoski:
So it's possible to use a reverse proxy with SpiderFoot running in a docker image.


smicallef:
Excellent stuff, merged into master. Thanks!

miaoski:
Thanks for merging!

PR #61, opened 2015-03-15
Codebase Maintenance

Gormogon:
Before I begin to describe this pull request, I'd like to thank you for all of the hard work you put into making such a great tool and making the source code publicly available.  Without your dedication, I would not have been able to save countless hours during the information gathering stage of the many pentests I have done.

------------------

What does this accomplish?
* Fair amount of simplifications
* Removal of trailing semicolons
* Declaration of file encoding (used the universal UTF-8) in accordance with PEP 263
* PEP 8 Comparison to 'None' fixes
* Loads of unused imports removed

This is the first part of my basic codebase maintenance. If this is accepted, I plan to bring the project as close to PEP compliance as possible (MANY PEP 8 compliance changes can be made).

Please let me know if there are any changes you would like me to make to this request.

------------------

References:
https://www.python.org/dev/peps/pep-0008/#programming-recommendations
https://www.python.org/dev/peps/pep-0263/

### DIFF
--- a/ext/adblockparser/parser.py
+++ b/ext/adblockparser/parser.py
@@ -98,7 +98,7 @@ class AdblockRule(object):
         else:
             self.raw_options = []
             self.options = {}
-        self._options_keys = frozenset(self.options.keys()) - set(['match-case'])
+        self._options_keys = frozenset(self.options.keys()) - {'match-case'}
 
         self.rule_text = rule_text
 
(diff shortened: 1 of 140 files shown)


smicallef:
Thanks a lot for taking the time to make all these fixes (and teaching me a thing or two about Python in the process.) The only issue is that I don't want to merge anything under ext/ because those are all external modules I ship with SpiderFoot, and may be replaced with newer versions as they are released (so your changes will be overwritten eventually.) Could you re-submit your pull request with everything excluding ext/?

Gormogon:
I should have been paying more attention...  Anyhow, all my changes to that directory have been reverted.  Had to create another commit to do so, not sure if there is a better way of doing this.

smicallef:
All merged, thanks a lot!

PR #63, opened 2015-03-15
Depreciation Resolution and More PEP 8

Gormogon:
This accomplishes three things:
* More PEP 8 Compliant.
* Removed redundant backslashes within brackets.
* Replaced all depreciated 'has_key()' instances with 'in'.

--------

'has_key()' has been depreciated in Python 2 and removed in Python 3. Additionally there is a performance gain by using 'in'.

To put this in perspective:
```
$ python2 -mtimeit -s'd=dict.fromkeys(range(99))' '12 in d'
10000000 loops, best of 3: 0.0339 usec per loop
$ python2 -mtimeit -s'd=dict.fromkeys(range(99))' 'd.has_key(12)'
10000000 loops, best of 3: 0.116 usec per loop
```

Due to the quantity of 'has_key()' instances that were replaced, this may provide a serious performance increase, or at least some sort of performance increase.

--------

References:
https://docs.python.org/2/library/stdtypes.html#dict.has_key

### DIFF
--- a/modules/sfp_accounts.py
+++ b/modules/sfp_accounts.py
@@ -85,8 +85,7 @@ class sfp_accounts(SpiderFootPlugin):
                                    )
 
             if res['content'] is None:
-                self.sf.debug("Unable to check the status of account " + name \
-                              + " on " + site)
+                self.sf.debug("Unable to check the status of account " + name + " on " + site)
             else:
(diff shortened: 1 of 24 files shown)


smicallef:
Thanks again, this is great!

Issue #70, opened 2015-07-21
Line Endings


felixrr:
Hey,

I recently cloned the spiderfoot repo and went to run it on my CentOS box.  Thinking it would be able to handle the bootstrap itself I just did ./sf.py, but this gave me line ending errors relating to "\r":

./sf.py: line 12: $'\r': command not found

Thinking this was just because it had been created on Windows I ran dos2unix over it.  Which, sort of worked...  It stopped me from getting the errors but i still couldn't run the application.  I eventually had a face-palm moment when I realised it just needed explicit execution;

python sf.py

However, I then get lots of weird hits on names and email addresses.  In particular lots of references to "5222" an "52225222" and "522252225222" and so on into infinity.

When I re-cloned this and ran it properly the first time, the problem doesn't seem to have come back.  Now, I'll be honest - no idea how dos2unix results in that weirdness but, I figure if the line endings can be fixed then this issue will never get noticed again.

Hope that helps - let me know if you need more detail.

Thanks

Felix

ccsplit:
Taking a look at the sf.put it looks like the issue is caused by not having the environment set using `#/usr/bin/env python ` which should resolve this issue on linux when running with just sf.py. Which is probably why you received the \r not a valid command because it was interpreting the file as an executable.

I will make a pull request later to fix this unless Steve does so before I get the chance.

smicallef:
Included in 2.5.1 which has just been released.

PR #72, opened 2015-08-06
Changes for Line endings

ccsplit:
Changed the Line endings for the sf.py file within the first commit and then added the `#!/usr/bin/env python` to let the shell know to setup the file for python environment when on linux for Issue #70 .


smicallef:
Merged, thanks!

PR #79, opened 2015-09-30
Fix Google Search

Wingless-Archangel:
This pull request is fix the bad request error as shown below:

```
Failed to fetch content from Google.

HTTP code 400 encountered for https://www.google.com/search?q=site:pastie.org +"agoda"&ie=utf-8&oe=utf-8&aq=t&rls=org.mozilla:en-US:official&client=firefox-a
```


smicallef:
Thanks a lot! Instead of merging those changes I decided to change sfp_pastebin into a module broadly covering paste sites - now sfp_pastes, covering the two paste sites you provided code for.

Regarding the google change you made, this seems like something that would result in a lot of CAPTCHAs given back from google?

Wingless-Archangel:
I tried to launch with this and the script didn't throw the captcha yet. I'm guessing that there's a sleep randomness in your script to avoid the bot from google. However, the spider still return junk which I think about spidering separately from the others as it is quite straightforward than ordinary spidering.

FYI: @shellcodenoob is my friend not me

smicallef:
No longer needed as we now use the APIs.

PR #84, opened 2015-11-07
Added RIPE Person Scan module

HackershubNL:
Module scans for person details by checking which person registered the
IP address

### DIFF
--- a/modules/sfp_registrar.py
+++ b/modules/sfp_registrar.py
@@ -1,0 +1,155 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         sfp_shodan
+# Purpose:      Query SHODAN for identified IP addresses.
+#
+# Author:      Steve Micallef <steve@binarypool.com>
+#
+# Created:     19/03/2014
+# Copyright:   (c) Steve Micallef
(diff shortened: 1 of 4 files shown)


smicallef:
Thanks! Just a few points:

- What was actually changed in sfdb.py? I don't see anything material.
- I'm not keen to include sfp_registrar as it's very specific to .nl. Can this be extended to other TLDs? Also it has a third party dependency on PhantomJS and only works on Linux. Have you looked at the sfp_whois module?
- Does sfp_ripe add anything beyond sfp_ir? sfp_ir already uses the RIPE API and is designed to be extensible to eventually also use others of the same nature (e.g. ARIN)

HackershubNL:
Hmm those commits weren't supposed to be in the pull request so I'll redo the pull request. But to your other questions:
- I added the PHONENR option in sfdb.py as the ripe module also fetches the person's phone number but that entity wasn't in the database yet.
- And yes the registrar module is currently for the .nl TLD but planning to expand it, I'll do a pull request as soon as I got the major TLDs in it.

smicallef:
OK, no problem, thanks. It's important that the modules have no platform-specific dependencies, e.g. running external command-line tools.

PR #86, opened 2015-11-09
Re-added the RIPE module and the change to sfdb.py also added SSL support and Digest Authentication

HackershubNL:
Re-added the RIPE module, it scans RIPE using the REST API together with any IP addresses found to get personal data on the person who reserved the IP address. It is dependent on the change to sfdb.py, this adds a new entity: PHONENR.
Changes to the sf.py file:
- Added SSL support, by default it checks whether there is a cert and key file with the name spiderfoot present. If not it'll create a new self signed 2048 bit sha256 certificate. If you got your own certificate you can comment out line 214 and update lines 215-216 for the path to your certificate. This does add a new prerequisite: python-openssl.
- Added Digest Authentication, it's the best that was available within cherrypy. It reads the secrets file in the spiderfoot directory for the users, so change this because now it is admin admin. It it fails to read the secrets file it'll fallback to a default admin admin login.

### DIFF
--- a/modules/sfp_ripe.py
+++ b/modules/sfp_ripe.py
@@ -1,0 +1,147 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         sfp_ripe
+# Purpose:      Query the RIPE REST API to find the person who registered the IP address
+#
+# Author:      Nander Hokwerda <nander@hackershub.co>
+#
+# Created:     7/11/2015
+# Copyright:   (c) Nander Hokwerda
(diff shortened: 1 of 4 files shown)


smicallef:
These are some great contributions, thanks a lot! Sorry for the delay responding.. been focusing on the HX beta. Regarding the RIPE module, sfp_ir already queries RIPE, so could your functionality be incorporated there? I want to avoid too many modules with overlapping data sources / functionality. Plus eventually I would like to add other data sources of a similar nature (e.g. ARIN).

Issue #83, opened 2015-11-06
Move downloads from SourceForge to Github


kruug:
Due to recent malicious actions by SourceForge, I would recommend moving all downloads to your GitHub release page.

smicallef:
OK, will do.

AntonioMeireles:
@smicallef - also, while on this, FWIW, no release post 2.5.1 is properly tagged in the source git tree here on github which is kind of... sad.

smicallef:
Ah crap, you're right! I had a script to do this automatically with each release but seems to have broken a couple of releases back. I'll tag the current codebase as 2.6.1 and ensure every release is tagged going forward. Thanks!

AntonioMeireles:
@smicallef fwiw this can be closed now :-) (and thanks!)

PR #87, opened 2015-11-16
sanitize Dockerfile

AntonioMeireles:
(now that #83 is actually handled)  builds against latest release and container has half the size now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
 #
-# Spiderfoot Dockerfile 
+# Spiderfoot Dockerfile
 #
 # http://www.spiderfoot.net
 #
-# Written by: Michael Pellon <m@pellon.io>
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,52 +1,51 @@
...
-  python-pip  \
-  python-dev \
-  python-lxml \
-  swig \
-  --no-install-recommends \
-  && rm -rf /var/lib/apt/lists/* \
-  && pip install cherrypy lxml mako M2Crypto netaddr
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        curl git python-pip python-lxml python-m2crypto python-netaddr \
```
imaia: Not a good thing to have multiple installs per line. https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

PR #93, opened 2015-12-23
Intelligence use case + cymon + sans

cudeso:
Added Intelligence option for UseCase
Updated VirusTotal to contain more details
Updated Shodan to contain more details
Added Cymon module
Added SANSISC module


smicallef:
Very cool contribution, thanks a lot!

Regarding Cymon and SANSISC, could they be incorporated into the sfp_malcheck module? That way it sticks with the model currently in place.

And as for the Intelligence use case, that is actually what the Investigate use case is for - what is the main difference you see?

cudeso:
Hi,

Some of the investigate modules (cookie, pageinfo) interact with the search subject, for certain domains/ip's that something I absolutely want to avoid. Proxying traffic only partly solves this because the subject still gets a hint of the connection. Hence the Intelligence-UC.

Regarding Cymon and Sans: it makes sense to include them with malcheck but
- I like to include all available info (geoinfo, extended subnet, contact details) from these sources (also think Xforce, OTX), even if this means double info; this can help to detect if sources return different info
- the malcheck module is becoming quite big, adding all the new different options for would introduce more complexity
- by using a module-setup I can run the searches without changing the settings, merely picking the correct modules to get the requested info

Including the basic checks (cymon and others) in the malcheck module would make sense though. I'll add them and see how it goes.

kr,

koen

smicallef:
OK, I see.. the Intelligence use case makes more sense now, I guess I will come up with a name for this use case vs. the other so there is a clearer distinction.

You make some good points regarding cymon and sans.. I wonder whether it makes sense to break apart sfp_malcheck into smaller pieces. Perhaps one module just for "dumb" lists and then specific modules for the more intelligence sources. The more I think about it, the more that makes sense, so I would say any sources that provide only malicious vs. not malicious information can stay in sfp_malcheck, and other richer sources will have their own modules. This means cymon and sans can stay as you originally intended - as separate modules. It will take me some time to adjust the others.

OK?

cudeso:
Sounds good. I can help with breaking-apart malcheck.

PR #100, opened 2016-04-15
PassiveTotal integration, using passivetotal pypi package

johannestaas:
http://passivetotal.org/

Hi! I'm on the passivetotal team at RiskIQ and I've been working on an integration with spiderfoot.
I'm not sure what the best way to go about adding an official is, but here's a patch to pull it into the modules subdir if this is the way to go about it.

This requires the passivetotal pypi package and it should already be set up with "pt-config setup"

This will give the functionality to query the PT API with IPs and Domains and generate new data through the passivedns, enrichment-osint and whois API calls.

Let me know if there's any reason not to pull it into master and if I should go about this another way. Thanks!

### DIFF
--- a/modules/sfp_passivetotal.py
+++ b/modules/sfp_passivetotal.py
@@ -131,6 +131,7 @@ class sfp_passivetotal(SpiderFootPlugin):
         ''' Performs setup of the module '''
         self.sf = sfc
         self.results = dict()
+        self.dedupe = set()
         self.client = PTClient()
         if self.client is None:
             self.sf.error('No PassiveTotal clients available, please configure '
@@ -157,6 +158,8 @@ class sfp_passivetotal(SpiderFootPlugin):
                 'SSL_CERTIFICATE_RAW']
 
     def handlePDNS(self, q):
+        if ('pdns', q) in self.dedupe:
+            return
         pdns_response = self.client.get_passive_dns(query=q)
         pdns_results = pdns_response.get('results', [])
         results = []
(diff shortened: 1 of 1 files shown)


9b:
@johannestaas could we pass "headers" with "PT-INTEGRATION" set to "Spiderfoot v1.0.0" on any request instance? Doing so allows us to track any issues that arise from our deployed module.

smicallef:
First of all, thanks for the contribution! Just a few questions/comments before I work on the merge:
- I see that the module requires a PassiveTotal "client" to be running? SpiderFoot is intended to not have any platform-specific dependency, so is your client able to run on Windows as well as Linux?
- I see no mention of an API key anywhere - is this not required?
- Why does the module watch for all events but then attempt to strip out IP Addresses, etc. when it could just watch for those specific events from all modules? This would save a lot of parsing.
- As I understand PassiveTotal, a more effective approach for this module would be to watch for IP_ADDRESS, INTERNET_NAME, DOMAIN_NAME and send those to PT's API.
- Please be aware that I try to avoid duplication in data from different sources. There are exceptions (for example port banners from SHODAN as well as direct port scans), but this is only when the source really has some additional value to add. For instance I am not sure what value there is in obtaining Whois information from PT as well as directly from the whois server (with the whois server obviously being preferred as it is the horses mouth).

johannestaas:
+Brandon

Hey Steve,

Sorry, haven't checked my email in a while. I'll need to check in with
Brandon who wrote the python API for a few answers here.

   - I see no mention of an API key anywhere - is this not required?
   - I see that the module requires a PassiveTotal "client" to be running?
   SpiderFoot is intended to not have any platform-specific dependency, so is
   your client able to run on Windows as well as Linux?

   I need to check whether Windows compatibility was tested. I haven't
   tested that personally. The client is installed via `pip install
   passivetotal` and `pt-config setup` handles setting up the credentials
   which get installed to ~/.config/passivetotal/api_config.json . The API key
   will end up there.

   - Why does the module watch for all events but then attempt to strip out
   IP Addresses, etc. when it could just watch for those specific events from
   all modules? This would save a lot of parsing.

   Probably just a bad design decision on my part. The passivetotal api has
   different endpoints and I shoved all of that into one class when it should
   probably be split into PDNS, Whois, SSLCert classes which take their
   respective types.

   - As I understand PassiveTotal, a more effective approach for this
   module would be to watch for IP_ADDRESS, INTERNET_NAME, DOMAIN_NAME and
   send those to PT's API.

   We'd probably want those with their MALICIOUS_ counterparts as well,
   maybe AFFILIATE_. From the look of other modules, maybe the NETBLOCK ones
   as well, and iterate across IPs with netaddr?

   - Please be aware that I try to avoid duplication in data from different
   sources. There are exceptions (for example port banners from SHODAN as well
   as direct port scans), but this is only when the source really has some
   additional value to add. For instance I am not sure what value there is in
   obtaining Whois information from PT as well as directly from the whois
   server (with the whois server obviously being preferred as it is the horses
   mouth).

   You make a good point about the whois implementation. There might be
   some added value in what PT provides, maybe more so in the future. Brandon
   might better answer this.

Cheers! Thanks for taking a detailed look.
Johan

bitvijays:
@johannestaas Hey Johan. Hope you are doing well. Is there any progress on this. I am really looking forward for this to integrate with spiderfoot :)

smicallef:
This is intentionally not merged due to the issues I raised earlier. I will eventually write a plug-in for PT, but using their REST API instead of the Python module.

PR #149, opened 2017-08-28
Add a module to detect SRV records

mscherer:
This PR add a module that will try to find servers using SRV records. This can be used to find AD servers or similar interesting system for internal pentesting, or external target (such as SIP, XMPP, etc) for exteranl intelligence gathering.

It also add a new record, both for UX and to potentially later be consumed by others modules.


smicallef:
Great, thanks for contributing this! Will merge right now..

PR #191, opened 2018-07-10
Redirect to docroot instead of "/"

BChip:
After scan deletions, redirect the user to docroot instead of "/".


smicallef:
Thanks!

PR #197, opened 2018-09-29
Add Skymem module

bcoles:


bcoles:
Added pagination.

Emails are scraped from the search results page. In the event there are no additional pages of results, the module terminates here.

If there are additional pages of results, the unique identifier for the domain is extracted, and unique email addresses are retrieved from the first 20 pages of results.

In the event there are less than 20 pages, on each page a check is performed to see if we're on the last page of results, and if so, the loop is terminated.

PR #198, opened 2018-09-29
Add EmailFormat module

bcoles:


bcoles:
I use it and find it useful. Sometimes the data is old.

Agreed, the sample data should be filtered.

PR #199, opened 2018-10-07
Add sfp_myspace module

bcoles:


smicallef:
I might be doing something wrong but I don't get any data out of this module using some test accounts that appear when I perform a search. For instance I found some test account here, which returns results, but nothing gets reported by the module: https://myspace.com/search/people?q=ad.test4.0@gmail.com

bcoles:
```
# The first result is the closest match, but whether it's an exact match is unknown.
# As such, we check for email address as name, at the risk of missed results.
```

Your example account has a username in the form of `<email> <email>` so it does't match.

```html
<a href="/113984082">ad.test4.0@gmail.com ad.test4.0@gmail.com</a></h6>
```

bcoles:
Fixed. The module now checks whether the user's name starts with the specified email address.

PR #200, opened 2018-10-08
Add Flickr module

bcoles:
Add Flickr module


modules/sfp_flickr.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if eventData in self.results:
+            return None
+        else:
+            self.results.append(eventData)
+
+        self.sf.debug("Received event, " + eventName + ", from " + srcModuleName)
+
+        if not eventName == "DOMAIN_NAME":
+            return None
```
smicallef: Not really necessary, because the module will only receive events it has expressed interest in from watchedEvents().

PR #204, opened 2018-10-13
Return resolved IP address

### LINKED ISSUE #203: sfp_opendns does not return resolved hosts
The [sfp_opendns](https://github.com/smicallef/spiderfoot/blob/master/modules/sfp_opendns.py) module resolves hosts with both the local resolver and the OpenDNS resolver.

The module only returns `MALICIOUS_*` events with a `Blocked by OpenDNS` message, and only if the host is in fact blocked by OpenDNS.

The resolved IP addresses are never returned.

bcoles:
Return resolved IP address from sfp_opendns module. Fix #203


smicallef:
I'm not so keen to have this module return resolved hosts, as that job should be delegated to sfp_dnsresolve, which also does wildcard DNS checks.

PR #212, opened 2018-10-18
Add sfp_portscan_udp module

bcoles:


smicallef:
I would rather not rely on scraping for something where the user would expect the port scan to be performed directly. Better would be to incorporate the functionality in the Hackertarget module, and hidden behind a configuration option which is off by default

bcoles:
> I would rather not rely on scraping for something where the user would expect the port scan to be performed directly.

Whose to say users expect the port scan to be performed directly?

Users familiar with [scanless](https://github.com/vesche/scanless) may have expectations to the contrary.

The module description doesn't imply a direct port scan.

bcoles:
Added port scan functionality to sfp_hackertarget module in #217. Closed.

PR #214, opened 2018-10-20
Add sfp_company module to Passive category

bcoles:


modules/sfp_company.py
```diff
@@ -19,7 +19,7 @@
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
 
 class sfp_company(SpiderFootPlugin):
-    """Company Names:Footprint,Investigate,Passive:Content Analysis::Identify company names in any obtained data."""
+    """Company Names:Passive,Footprint,Investigate,Passive:Content Analysis::Identify company names in any obtained data."""
```
smicallef: Passive is already there :)

bcoles: Welp. You're right. I didn't even notice. It looks like there's a bug lurking here.

Despite `Found company name` appearing in the logs, and other modules declaring receipt of a `COMPANY_NAME` event, the company name does not appear in the `Browse` tab, nor when exporting the results as CSV.

smicallef: Make sure you're not using an old version of the database..

bcoles: Fair call. That fixed it. Thanks.

PR #216, opened 2018-10-20
Add sfp_opencorporates module

bcoles:


modules/sfp_opencorporates.py
```diff
@@ -0,0 +1,131 @@
...
+        for opt in userOpts.keys():
+            self.opts[opt] = userOpts[opt]
+
+    # What events is this module interested in for input
+    def watchedEvents(self):
+        return ["COMPANY_NAME"]
+
+    # What events this module produces
+    def producedEvents(self):
+        return ["AFFILIATE_COMPANY_NAME", "GEOINFO"]
```
smicallef: Can we get a bit more, like the human names of company directors?

modules/sfp_opencorporates.py
```diff
@@ -0,0 +1,131 @@
...
+import re
+import urllib
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
+class sfp_opencorporates(SpiderFootPlugin):
+    """OpenCorporates:Passive,Footprint,Investigate:Search Engines::Look up company information from OpenCorporates."""
+
+    # Default options
+    opts = {
+        'confidence': 100
```
smicallef: What about the ability for the user to enter in their own API key?

modules/sfp_opencorporates.py
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    # Option descriptions
+    optdescs = {
+        'confidence': "Confidence that the search result objects are correct (numeric value between 0 and 100)."
```
smicallef: It's not a field that's used today, so not sure it's worth adding.

modules/sfp_opencorporates.py
```diff
@@ -0,0 +1,131 @@
...
+            self.sf.debug("Found no results for " + eventData)
+            return None
+
+        # Check for a match
+        for company in data['results']['companies']:
+            if not eventData.lower() == company['company']['name'].lower():
+                continue
+
+            # Extract registered address
+            location = company['company']['registered_address_in_full']
```
smicallef: Will this field always exist? Better to use .get() to avoid an exception here.

modules/sfp_opencorporates.py
```diff
@@ -0,0 +1,131 @@
...
+                            country = company['company']['registered_address']['country']
+                            if not location.endswith(country):
+                                location = location + ", " + country
+
+                    self.sf.info("Found company address: " + location)
+                    e = SpiderFootEvent("GEOINFO", location, self.__name__, event)
+                    self.notifyListeners(e)
+
+            # Extract previous company names
+            previous_names = company['company']['previous_names']
```
smicallef: A previous name isn't an affiliate though?

bcoles: True, but that seemed to fit with the existing paradigm. Similar to how the `AFFILIATE_` prefix is used elsewhere.

I also wasn't sure if you would be keen to have this module report previous names at all. For example, no modules other than the `sfp_names` module are permitted to return a `HUMAN_NAME`. I wasn't sure if the same applied to company names.

bcoles: `RAW_RIR_DATA` instead, perhaps?

modules/sfp_opencorporates.py
bcoles: What do you mean? This field is used.

modules/sfp_opencorporates.py
bcoles: This can be retrieved, at the cost of an extra HTTP request. Only some companies have listed officers.

The problem is, even when `confidence` is set to `100`, it's still possible that two companies in two different countries will have the exact same name.

The module fires on `COMPANY_NAME` event, but without any associated location data. This could result in reporting a bunch of `HUMAN_NAME` objects for a completely different company.

bcoles: Updated to retrieve officer names.

modules/sfp_opencorporates.py
bcoles: Updated to support API keys

modules/sfp_opencorporates.py
bcoles: Updated to use `get()` for retrieval of JSON properties.

modules/sfp_opencorporates.py
bcoles: Updated to use `COMPANY_NAME`.

bcoles:
Suggested changes have been implemented.

PR #218, opened 2018-10-26
Add sfp_spyonweb module

bcoles:


sfdb.py
```diff
@@ -215,6 +215,7 @@ class SpiderFootDb:
         "INSERT INTO tbl_event_types (event, event_descr, event_raw, event_type) VALUES ('URL_UPLOAD_HISTORIC', 'Historic URL (Accepts Uploads)', 0, 'DESCRIPTOR')",
         "INSERT INTO tbl_event_types (event, event_descr, event_raw, event_type) VALUES ('USERNAME', 'Username', 0, 'ENTITY')",
         "INSERT INTO tbl_event_types (event, event_descr, event_raw, event_type) VALUES ('VULNERABILITY', 'Vulnerability in Public Domain', 0, 'DESCRIPTOR')",
+        "INSERT INTO tbl_event_types (event, event_descr, event_raw, event_type) VALUES ('WEB_ANALYTICS_ID', 'Web Analytics', 0, 'DATA')",
```
smicallef: I think this better fits as an entity, as it's an identifier for something (i.e. an account owned by a company) and not just raw data.

modules/sfp_spyonweb.py
```diff
@@ -0,0 +1,367 @@
...
+            self.sf.info("No results found for " + qry)
+            return None
+
+        try:
+            data = json.loads(res['content'])
+        except Exception as e:
+            self.sf.debug("Error processing JSON response.")
+            return None
+
+        if not data['status'] == 'found':
```
smicallef: Please use .get() in all the cases you are referring to dictionary keys, as it will possibly raise exceptions when the 'status' key isn't set. Also in other cases the key may be set but may be zero, which will result in a different exception.

sfdb.py
bcoles: Updated

modules/sfp_spyonweb.py
bcoles: I've changed to `get()` in the instances where I think this is reasonable.

smicallef: The other should still have it. For instance, below on 108 you correctly check if result is a key in data, but then further down on 112 you are assuming that data['result'] is not going to be None. If it is though, an exception will be raised and the whole scan will fail. Same for various others below.

modules/sfp_spyonweb.py
```diff
@@ -0,0 +1,367 @@
...
+                    return True
+        return False
+
+    # What events is this module interested in for input
+    def watchedEvents(self):
+        return ['IP_ADDRESS', 'INTERNET_NAME', 'DOMAIN_NAME', 'WEB_ANALYTICS_ID']
+
+    # What events this module produces
+    def producedEvents(self):
+        return ['CO_HOSTED_SITE', 'INTERNET_NAME', 'AFFILIATE_DOMAIN', 'WEB_ANALYTICS_ID']
```
smicallef: Should we maybe introduce a module that simply extracts analytics IDs from TARGET_WEB_CONTENT events? That way this module can also act on any analytics IDs extracted from web content taken from scanned sites.

bcoles: Yeah I was thinking the same thing. I can't think of any good reason to not implement something like that.

PR #220, opened 2018-10-28
Add sfp_openstreetmaps

bcoles:
Could be improved with some normalization of the `PHYSICAL_ADDRESS` prior to querying.

However, this would come at the expense of adding a library dependency. [libpostal](https://github.com/openvenues/libpostal) appears to be the best option.


modules/sfp_openstreetmaps.py
```diff
@@ -0,0 +1,123 @@
...
+# -------------------------------------------------------------------------------
+
+import json
+import re
+import time
+import urllib
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
+class sfp_openstreetmaps(SpiderFootPlugin):
+    """Open Street Maps:Footprint,Investigate,Passive:Real World::Retrieves latitude/longitude coordinates for physical addresses from Open Street Maps API."""
```
smicallef: should be OpenStreetMap, as that's the official name of the service?

modules/sfp_openstreetmaps.py
```diff
@@ -0,0 +1,123 @@
...
+            self.sf.debug("Found no results for " + eventData)
+            return None
+
+        self.sf.info("Found " + str(len(data)) + " matches for " + eventData)
+
+        for location in data:
+            lat = str(location.get('lat'))
+            lon = str(location.get('lon'))
+
+            if not lat or not lon:
```
smicallef: If the previous `.get()` calls return None, they will be converted to the string `None` and therefore this `if` will never return `True`.

PR #221, opened 2018-10-28
Add sfp_h1nobbdde

RootUp:
Added support for `sfp_h1nobbdde` which query the [unofficial HackerOne ](http://h1.nobbd.de/)disclosure timeline database  to see if our target appears.

### DIFF
--- a/modules/sf_h1nobbdde.py
+++ b/modules/sf_h1nobbdde.py
@@ -16,7 +16,7 @@ import json
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
 
 class sfp_h1nobbdde(SpiderFootPlugin):
-     """the unofficial HackerOne:Footprint,Investigate,Passive:Leaks, Dumps and Breaches::Check external vulnerability scanning/reporting service h1.nobbd.de to see if the target is listed."""
+     """Unofficial HackerOne:Footprint,Investigate,Passive:Leaks, Dumps and Breaches::Check external vulnerability scanning/reporting service h1.nobbd.de to see if the target is listed."""
 
 
 
@@ -48,7 +48,9 @@ class sfp_h1nobbdde(SpiderFootPlugin):
 
     # What events this module produces
     def producedEvents(self):
-         return ["*"]
+        ret = ["VULNERABILITY"]
+
+        return ret
 
     # Query h1.nobbd.de
     def queryOBB(self, qry):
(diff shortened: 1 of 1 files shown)


modules/sf_h1nobbdde.py
```diff
@@ -0,0 +1,99 @@
...
+
+import sys
+import time
+import datetime
+import re
+import json
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
+class sfp_h1nobbdde(SpiderFootPlugin):
+    """The unofficial HackerOne disclosure timeline check external vulnerability scanning/reporting service from h1.nobbd.de to see if the target is listed."""
```
smicallef: This needs to be a specific format - please take a look at other modules for examples.

modules/sf_h1nobbdde.py
```diff
@@ -0,0 +1,99 @@
...
+        self.results = dict()
+
+        # Clear / reset any other class member variables here
+        # or you risk them persisting between threads.
+
+        for opt in userOpts.keys():
+            self.opts[opt] = userOpts[opt]
+	
+    def watchedEvents(self):
+        return ["INTERNET_NAME"]
```
smicallef: Do we need to look for all hostnames or would `DOMAIN_NAME` be enough here?

modules/sf_h1nobbdde.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # What events this module produces
+    def producedEvents(self):
+        ret = ["VULNERABILITY"]
```
smicallef: Just do a return of this instead of allocating a variable.

modules/sf_h1nobbdde.py
```diff
@@ -44,13 +44,11 @@ def setup(self, sfc, userOpts=dict()):
...
     def watchedEvents(self):
-        return ["INTERNET_NAME"]
+        return ["DOMAIN_NAME"]
 
     # What events this module produces
     def producedEvents(self):
-        ret = ["VULNERABILITY"]
-
-        return ret
+         return ["*"]
```
smicallef: The module needs to return `VULNERABILITY`, you had it right there the first time.

modules/sf_h1nobbdde.py
```diff
@@ -16,7 +16,7 @@
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
 
 class sfp_h1nobbdde(SpiderFootPlugin):
-    """The unofficial HackerOne disclosure timeline check external vulnerability scanning/reporting service from h1.nobbd.de to see if the target is listed."""
+     """the unofficial HackerOne:Footprint,Investigate,Passive:Leaks, Dumps and Breaches::Check external vulnerability scanning/reporting service h1.nobbd.de to see if the target is listed."""
```
smicallef: `the` seems unnecessary? I try to keep the name short so that it doesn't mess with the UI too much.

smicallef:
Great job! Thanks for the contribution!

PR #232, opened 2018-12-01
Add sfp_neutrinoapi

bcoles:


modules/sfp_neutrinoapi.py
```diff
@@ -0,0 +1,164 @@
...
+        'api_key': "NeutrinoAPI API key.",
+        'timeout': "Query timeout, in seconds."
+    }
+
+    results = dict()
+    errorState = False
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.__dataSource__ = "NeutrinoAPI"
```
smicallef: Just FYI for the future, you only need to set this if you intend to override the module name, otherwise the module name is used by default as the data source.

smicallef:
Nice stuff, thanks @bcoles !

PR #236, opened 2019-01-02
Add bucket suffixes

TheCrott:
Based on what I found and other reports, I add some s3 suffix like prod, production, staging, etc


bcoles:
How far do you want to go with it?

`stg`, `uat`, `test`, `prd`, `preprd`, `preprod` are all good candidates too.

PR #244, opened 2019-03-04
Initial support for ReconJSON

bcoles:
Add support for exporting as ReconJSON

Draft / request for comments.

I'd like to see support for ReconJSON added eventually. This approach may not be the best approach. On the other hand, it may be better than nothing. In the long term, the data for each host should should probably be merged into a `dict` for each host, before dumping as JSON.

---

Supports only hosts. Does not merge results, resulting in duplicate `Hosts`. This is permitted by the specification, but probably not ideal. This approach will likely prove to be problematic in the future, with the addition of support for [Services](https://github.com/ReconJSON/ReconJSON/blob/master/Service.md) objects. Expanding upon the current approach to support exporting services would result in one `Host` entry for each open port.

Exports all data types relating to [Hosts](https://github.com/ReconJSON/ReconJSON/blob/master/Host.md) in one way or another (includes co-hosted sites, affiliates, similar domains, etc).

This does not include host groups, such as netblocks, ASN, subnets, etc.

Does not export anything related to [DNS](https://github.com/ReconJSON/ReconJSON/blob/master/DNS.md) (mx records, name servers, etc).

Does not export anything related to [Services](https://github.com/ReconJSON/ReconJSON/blob/master/Service.md) (open ports, banners, software, etc).

### DIFF
--- a/dyn/scanlist.tmpl
+++ b/dyn/scanlist.tmpl
@@ -117,8 +117,12 @@
                     var efr = document.getElementById('exportframe');
                     if (type == "gexf") {
                         efr.src = '${docroot}/scanvizmulti?ids=' + ids.join(',');
-                    } else {
+                    } else if (type == "csv") {
                         efr.src = '${docroot}/scaneventresultexportmulti?ids=' + ids.join(',');
+                    } else if (type == "reconjson") {
+                        efr.src = '${docroot}/scanexportreconjsonmulti?ids=' + ids.join(',');
+                    } else {
+                        console.log("Invalid export type");
                     }
                     $("#loader").fadeOut(500);
                 }
@@ -168,7 +172,9 @@
                             buttons += "<button class='btn dropdown-toggle btn-success' data-toggle='dropdown'><span class='caret'></span></button>";
                             buttons += "<ul class='dropdown-menu'>";
                             buttons += "<li><a href='javascript:exportSelected(\"csv\")'>CSV</a></li>";
-                            buttons += "<li><a href='javascript:exportSelected(\"gexf\")'>GEXF</a></li></ul>";
+                            buttons += "<li><a href='javascript:exportSelected(\"gexf\")'>GEXF</a></li>";
+                            buttons += "<li><a href='javascript:exportSelected(\"reconjson\")'>ReconJSON</a></li>";
+                            buttons += "</ul>";
                             buttons += "</div>";
 
                             buttons += "<div class='btn-group pull-right'>";
--- a/sfwebui.py
+++ b/sfwebui.py
@@ -175,6 +175,45 @@ class SpiderFootWebUi:
 
     scansearchresultexport.exposed = True
 
+    # Export results from multiple scans in ReconJSON format
+    def scanexportreconjsonmulti(self, ids):
+        types = list()
+        dbh = SpiderFootDb(self.config)
+        sf = SpiderFoot(self.config)
+        data = list()
+
+        for id in ids.split(','):
+            data = data + dbh.scanResultEvent(id, filterFp=True)
+
+        host_events = [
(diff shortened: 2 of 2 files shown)


smicallef:
Thanks @bcoles, really appreciated as always. My only concern here is that ReconJSON doesn't appear to be widely adopted/recognised and the idea with exports is to include everything visible in that section. I think it might lead to confusion for those who don't know the format and expect to get everything exported, but instead only getting a (fairly small) subset of the data SpiderFoot has collected.

bcoles:
tl/dr: Perhaps, the best approach is to write a third-party tool which converts SpiderFoot CSV to ReconJSON. How do you feel about me implementing a straight up one-for-one JSON export feature?

---

> My only concern here is that ReconJSON doesn't appear to be widely adopted/recognised 

ReconJSON is a new "standard" - around 6 months old - and has yet to be adopted by [m]any projects.

`ReconJSON is a project dedicated to creating a flexible and consistent JSON format across popular recon tools.`

If ReconJSON can't be reasonably used as an export feature, then this is an existential failure of the ReconJSON project.

There is [discussion](https://github.com/ReconJSON/ReconJSON/issues/13) around reviewing SpiderFoot's data types with the intention of implementing these in ReconJSON. However, there likely will never be one-for-one parity. Progress in this area may or may not result in improvements to SpiderFoot, and will likely bring more eyes on both projects. The best case scenario would be SpiderFoot inspiring and influencing ReconJSON development, resulting in close to one-to-one conversion.

For example, one issue with the standard which became immediately obvious was the inability to add any form of `source data` attribute, indicating from where the data was acquired. Personally, I consider this to be a shortcoming of the ReconJSON standard. A similar issue is the lack of timestamps, which has been raised and [discussed](https://github.com/ReconJSON/ReconJSON/issues/10).

I can understand not wanting to be the guenie pig for ReconJSON development. None the less, for ReconJSON to be successful, one project must adopt it first, and deal with the associated teething problems.

If recon tools don't adopt exporting to ReconJSON format, the alternative is development of a bunch of parsers, which take output from various recon tools and convert it to ReconJSON for the purposes of then importing this data into other recon tools. Unfortunately, this is a possible outcome for the format.

For example, I can't ever see nmap implementing a ReconJSON output format, but I can certainly see third-party tools/parsers developed to convert nmap output to ReconJSON.


Note: I'm in no way affiliated with the ReconJSON project. I just like the idea.

> and the idea with exports is to include everything visible in that section. I think it might lead to confusion for those who don't know the format and expect to get everything exported, but instead only getting a (fairly small) subset of the data SpiderFoot has collected.

That's fair enough. On the other hand, I'm fairly certain GEXF export format does not export all data.

Similarly, I don't expect to get lossless image quality when I save an SVG image as JPG. I expect to get a JPG image in all its unholy compressed glory.

Confusion will decrease roughly proportionally to the increase of ReconJSON popularity; presuming, of course, that ReconJSON becomes popular.

These concerns could be partially mitigated with a simple modification to the export label. ie, `Export -> ReconJSON (Hosts)` would clearly imply that only host data is exported.

> only getting a (fairly small) subset of the data SpiderFoot has collected

The natural progression is to include [Service](https://github.com/ReconJSON/ReconJSON/blob/master/Service.md) related data types such as `SOFTWARE`, `OPEN_[UDP|TCP]_PORT`, `TCP_PORT_OPEN_BANNER`, etc. Granted, that still leaves a large amount of data ignored during export.

bcoles:
I've opened a PR to add JSON exporting functionality #246

smicallef:
I've merged the generic JSON export feature. Let's wait just a little while longer until ReconJSON matures (just a bit) and then we can come back to it. The more export/import functionality SpiderFoot has, the better. I just don't want to be in a position where we have to make continual updates to the export feature because the format is evolving so quickly.

PR #248, opened 2019-03-25
Catch backspace, instead of navigate go to main results page

HackershubNL:
When using the backspace key when navigating through the results it will navigate back to the scans overview. This catches the button press and returns the user to the scan results overview.


bcoles:
This should probably be resolved by using `history.pushState()`, rather than hijacking the backspace key, so as to also resolve the issue for use of the `back` button within the browser (for which the backspace key is a shortcut.)

smicallef:
Thanks for this, but this has fallen pretty far behind and includes changes not related to the PR's stated intent.

PR #257, opened 2019-04-11
Dockerize current directory, instead of "https://github.com/smicallef/spiderfoot/archive/master.tar.gz"

ikp4success:


Dockerfile
```diff
@@ -15,7 +16,9 @@
 
 # Pull the base image.
 FROM alpine:latest
-COPY requirements.txt .
+WORKDIR /home/spiderfoot
+COPY . .
+ENV SPIDERFOOT_VERSION 3.0.0
```
smicallef: I guess if we're going with using the Dockerfile to always produce from what's local instead of the master branch, we should remove this version seeing as we won't know what the verison will be.

steve-bate:
#574 also includes this feature.

Dockerfile
leotrubach: @ikp4success why are you removing the `COPY requirements.txt .`?
https://stackoverflow.com/a/34399661/1831675

leotrubach:
So the docker file should be the following (to avoid building images every time):
1) install apt packages
2) copy only requirements.txt to work dir
3) pip install requirements
4) copy the rest

that will skip pip install step if requirements.txt is not changed.

steve-bate:
The current Dockerfile on master does what you are describing, right?

leotrubach:
> The current Dockerfile on master does what you are describing, right?

I think not. 
It does
1) copy requirements.txt .
2) install apk packages, *pip install*, add users, make data dir, 
...
3) copy . .

I suggest that it is changed to
1) install apk packages, add users, make data dir, 
2) copy requirements.txt .
3) pip install
4) copy . .

This will allow to repeat only steps 2-3-4 when some new package is added to requirements.txt thus we get quicker subsequent builds.
Well requirements.txt is updated rarely (due to philosophy of spiderfoot), so maybe not a very big deal.

steve-bate:
I understand now and I agree that what you're suggesting would build faster when a new package is added to requirements.txt. However, as you said, that doesn't happen very often.

smicallef:
Thanks for the submission, but this is now surpassed by https://github.com/smicallef/spiderfoot/pull/1005.

PR #258, opened 2019-04-11
sfp_malwarepatrol fix

ikp4success:


modules/sfp_malwarepatrol.py
```diff
@@ -45,30 +47,44 @@ def watchedEvents(self):
...
+            "spamassassin",
+            "smoothwall"
+        ]
+
+        for block_name in block_list:
+            data = dict()
+            url = "http://lists.malwarepatrol.net/cgi/getfile?receipt={}&product=32&list={}"
+            url = url.format(self.opts['api_key'], block_name)
+
+            data['content'] = self.sf.cacheGet("sfmalwarepatrol", 72)
```
smicallef: The Cache ID will need to be different for each block list, otherwise only the last in the list will be the one cached.

modules/sfp_malwarepatrol.py
```diff
@@ -45,30 +47,44 @@ def watchedEvents(self):
...
-                self.sf.error("Unable to fetch " + url, False)
-                return None
-            else:
-                self.sf.cachePut("sfmalwarepatrol", data['content'])
+                data = self.sf.fetchUrl(url, useragent=self.opts['_useragent'])
+                if data['content'] is None:
+                    self.sf.error("Unable to fetch " + url, False)
+                    return None
+                else:
+                    self.sf.cachePut("sfmalwarepatrol", data['content'])
```
smicallef: Previous comment applies here too.

modules/sfp_malwarepatrol.py
```diff
@@ -11,8 +11,10 @@
 # Licence:     GPL
 # -------------------------------------------------------------------------------
 
-import json
-from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+from sflib import (
+    SpiderFootPlugin,
+    SpiderFootEvent
+)
```
smicallef: I'd prefer you stuck with the convention used in other modules.

smicallef:
Thanks!

smicallef:
Actually I just realised that these are not actually different feeds, but just different formats of the same data. So, I'll revert this change. Sorry.

PR #261, opened 2019-04-29
Add WEB_ANALYTICS_ID to sfp_shodan watched events

bcoles:
Add `WEB_ANALYTICS_ID` to `sfp_shodan` watched events.

Allows searching for similar sites using the same analytics IDs, such as Google analytics, by using the `http.html:` filter. Saves results as `SEARCH_ENGINE_WEB_CONTENT`.

Saving as `SEARCH_ENGINE_WEB_CONTENT` is not ideal, but this is currently what the module also does for `DOMAIN_NAME` events. Ideally, in the future, the response should be parsed and the appropriate events returned. I consider this outside the scope of this PR.

The results from this test are usually not as good as SpyOnWeb, however, it's nice to have options.

Also, **this PR ninja patches a bug** where the URL should use `&` instead of `?` which was causing the API to return `401`.

### DIFF
--- a/modules/sfp_shodan.py
+++ b/modules/sfp_shodan.py
@@ -136,8 +136,14 @@ class sfp_shodan(SpiderFootPlugin):
             self.notifyListeners(evt)
 
         if eventName == 'WEB_ANALYTICS_ID':
-            network = eventData.split(": ")[0]
-            analytics_id = eventData.split(": ")[1]
+            try:
+                network = eventData.split(": ")[0]
+                analytics_id = eventData.split(": ")[1]
+            except BaseException as e:
+                self.sf.error("Unable to parse WEB_ANALYTICS_ID: " +
+                              eventData + " (" + str(e) + ")", False)
+                return None
+
             rec = self.searchHtml(analytics_id)
 
             if rec is None:
--- a/modules/sfp_shodan.py
+++ b/modules/sfp_shodan.py
@@ -21,7 +21,7 @@ class sfp_shodan(SpiderFootPlugin):
 
     # Default options
     opts = {
-        "api_key": "",
+        'api_key': "",
         'netblocklookup': True,
         'maxnetblock': 24
     }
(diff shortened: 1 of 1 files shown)


modules/sfp_shodan.py
```diff
@@ -120,6 +135,18 @@ def handleEvent(self, event):
             evt = SpiderFootEvent("SEARCH_ENGINE_WEB_CONTENT", str(hosts), self.__name__, event)
             self.notifyListeners(evt)
 
+        if eventName == 'WEB_ANALYTICS_ID':
+            network = eventData.split(": ")[0]
+            analytics_id = eventData.split(": ")[1]
+            rec = self.searchHtml(analytics_id)
+
```
smicallef: Perhaps make this upper block a bit safer by catching an exception in case parsing fails?

modules/sfp_shodan.py
```diff
@@ -120,6 +135,18 @@ def handleEvent(self, event):
...
+            if rec is None:
+                return None
+
+            evt = SpiderFootEvent("SEARCH_ENGINE_WEB_CONTENT", str(rec), self.__name__, event)
```
smicallef: Can we return the content instead of wrapping it in `str()`? Doubt that would be unicode safe.

bcoles: Should the `'DOMAIN_NAME'` block also be updated?

This block is essentially copy pasta of existing code which also uses `str()`:


```py
        if eventName == "DOMAIN_NAME":
            hosts = self.searchHosts(eventData)
            if hosts is None:
                return None
            
            evt = SpiderFootEvent("SEARCH_ENGINE_WEB_CONTENT", str(hosts), self.__name__, event)
            self.notifyListeners(evt)

        if eventName == 'WEB_ANALYTICS_ID':
            network = eventData.split(": ")[0]
            analytics_id = eventData.split(": ")[1]
            rec = self.searchHtml(analytics_id)

            if rec is None:
                return None

            evt = SpiderFootEvent("SEARCH_ENGINE_WEB_CONTENT", str(rec), self.__name__, event)
            self.notifyListeners(evt)
            return None
```

modules/sfp_shodan.py
bcoles: Fixed. Untested. Copy pasta error handling from another module.

PR #271, opened 2019-05-12
Add sfp_darksearch

bcoles:


modules/sfp_darksearch.py
```diff
@@ -0,0 +1,126 @@
...
+        return ['DARKNET_MENTION_URL', 'DARKNET_MENTION_CONTENT', 'SEARCH_ENGINE_WEB_CONTENT']
+
+    # https://darksearch.io/apidoc
+    def query(self, qry, page):
+        params = {
+            'query': qry.encode('raw_unicode_escape'),
+            'page': str(page)
+        }
+
+        res = self.sf.fetchUrl("https://darksearch.io/api/search?" + urllib.urlencode(params),
```
smicallef: `urllib.urlencode()` is not Unicode safe afaik.

modules/sfp_darksearch.py
```diff
@@ -0,0 +1,126 @@
...
+import time
+import urllib
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
+class sfp_darksearch(SpiderFootPlugin):
+    """Darksearch:Footprint,Investigate:Search Engines::Search the Darksearch.io Tor search engine for mentions of the target domain."""
+
+    # Default options
+    opts = {
+        'max_pages': 20
```
smicallef: Would be reeeeally nice if you could also implement the `fetchlinks` functionality :)

modules/sfp_darksearch.py
bcoles: What's best practice here? The framework appears to take a few approaches.

This approach is copypasta from another module (`sfp_openstreetmap`), in which you added support for unicode in a413501e78e6d295d8bee546bed6201511a7061b by using `encode('raw_unicode_escape')`.

modules/sfp_darksearch.py
bcoles: Done

PR #275, opened 2019-05-25
Add sfp_numverify module

bcoles:
Depends on #274 to add `PROVIDER_TELCO` data type.


modules/sfp_numverify.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.sf.debug("Received event, " + eventName + ", from " + srcModuleName)
+
+        data = self.query(eventData)
+
+        if data is None:
+            self.sf.debug("No phone information found for " + eventData)
+            return None
+
+        evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef: Possible unicode issue here.

PR #276, opened 2019-05-25
Add phone number lookup to sfp_neutrinoapi module

bcoles:
Add phone number lookup to sfp_neutrinoapi module.

Also ninja patches an issues where attempts to store `RAW_RIR_DATA` were failing due to attempting to store the API response `data` as a `dict` rather than a `str`.

### DIFF
--- a/modules/sfp_neutrinoapi.py
+++ b/modules/sfp_neutrinoapi.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 # -------------------------------------------------------------------------------
-# Name:         sfp_neutrinoapi
-# Purpose:      SpiderFoot plug-in to search NeutrinoAPI for IP address info and
-#               check IP reputation.
+# Name:        sfp_neutrinoapi
+# Purpose:     SpiderFoot plug-in to search NeutrinoAPI for IP address info,
+#              check IP address reputation, and search for phone location.
 #
-# Author:      Brendan Coles <bcoles@gmail.com>
+# Author:      <bcoles@gmail.com>
 #
 # Created:     2018-11-30
-# Copyright:   (c) Brendan Coles 2018
+# Copyright:   (c) bcoles 2018
 # Licence:     GPL
 # -------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


modules/sfp_neutrinoapi.py
```diff
@@ -130,36 +136,59 @@ def handleEvent(self, event):
...
+            else:
+                if data.get('location') is not None and data.get('country') is not None:
+                    if data.get('location') == data.get('country'):
+                        location = data.get('location')
+                    else:
+                        location = data.get('location') + ', ' + data.get('country')
+
+                    evt = SpiderFootEvent("GEOINFO", location, self.__name__, event)
+                    self.notifyListeners(evt)
+                    evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef: Not sure if this might cause issues when `data` is unicode.

modules/sfp_neutrinoapi.py
```diff
@@ -130,36 +136,59 @@ def handleEvent(self, event):
...
+
+            data = self.queryIpBlocklist(eventData)
+
+            if data is None:
+                self.sf.debug("No IP blocklist results found for " + eventData)
+            else:
+                if data.get('is-listed'):
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", eventData, self.__name__, event)
+                    self.notifyListeners(evt)
+                    evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef: Possible unicode issue here too.

modules/sfp_neutrinoapi.py
```diff
@@ -130,36 +136,59 @@ def handleEvent(self, event):
...
+
+            data = self.queryHostReputation(eventData)
+
+            if data is None:
+                self.sf.debug("No host reputation results found for " + eventData)
+            else:
+                if data.get('is-listed'):
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", eventData, self.__name__, event)
+                    self.notifyListeners(evt)
+                    evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef: And here.

PR #285, opened 2019-05-28
Add sfp_callername

bcoles:


modules/sfp_callername.py
```diff
@@ -0,0 +1,119 @@
...
+
+        rep_good_match = re.findall(r'>SAFE.*?>(\d+) votes?<', html)
+        rep_bad_match = re.findall(r'>UNSAFE.*?>(\d+) votes?<', html)
+
+        if rep_good_match is not None and rep_bad_match is not None:
+            good_votes = int(rep_good_match[0])
+            bad_votes = int(rep_bad_match[0])
+
+            if bad_votes > good_votes:
+                evt = SpiderFootEvent('MALICIOUS_PHONE_NUMBER', eventData, self.__name__, event)
```
smicallef: Can we include a link to the callername website for this number like we do in other modules?

bcoles: Done

PR #303, opened 2019-06-05
Quote search query

### LINKED ISSUE #301: sfp_darksearch reduce false positives
sfp_darksearch currently searches without quotes. For example "John Doe" is searched as John Doe returning all the pages having John or Doe, that is a huge ton of unuseful data.
At least as an option, you should be able to issue "quoted" searches that seem to be supported by the search engine. This greatly reduces false, especially looking for people.

bcoles:
I haven't tested this, but it should fix #301

### DIFF
--- a/modules/sfp_darksearch.py
+++ b/modules/sfp_darksearch.py
@@ -52,7 +52,7 @@ class sfp_darksearch(SpiderFootPlugin):
     # https://darksearch.io/apidoc
     def query(self, qry, page):
         params = {
-            'query': qry.encode('raw_unicode_escape'),
+            'query': '"' + qry.encode('raw_unicode_escape') + '"',
             'page': str(page)
         }
 


LoZio:
It does, thank you. Also the search engine is smart enough to match "John Doe" against 
`<em>JOHN</em> <em>DOE</em>`
so it runs much better now.
Other enhancements would be:
- allow to choose if use quoted search or not, a checkbox in the settings of the module. I like this new way, other may like older "gimme back something" way. This should be a general "search engine module developement guideline" in my opinion even not all the (darknet) search engine currently support this.
- flag to allow to also search for swapped name/surname when looking for people. "John Doe" and "Doe John" may be present depending on the page content and sorting.

I'll try to learn using github/PR asap to be more useful and provide the changes instead of only asking :)

bcoles:
> 
> 
> It does, thank you. Also the search engine is smart enough to match "John Doe" against
> `<em>JOHN</em> <em>DOE</em>`
> so it runs much better now.

You'll probably find that the search engine is adding the `<em>` tags for highlighting purposes.

> Other enhancements would be:
> 
>     * allow to choose if use quoted search or not, a checkbox in the settings of the module. I like this new way, other may like older "gimme back something" way. This should be a general "search engine module developement guideline" in my opinion even not all the (darknet) search engine currently support this.
> 
>     * flag to allow to also search for swapped name/surname when looking for people. "John Doe" and "Doe John" may be present depending on the page content and sorting.
> 

I didn't particularly care about searching for `HUMAN_NAME` when I wrote the DarkSearch module. I only added this for consistency with the other darknet related modules.

I agree with adding a `quoted search` or `exact match` option to modules where the queried services supports such queries.

Any changes made related to formatting of `HUMAN_NAME` entities (ie, `first last`, `last, first`, etc) should be applied consistently. That is, to all darknet related modules which accept a human name, or all modules which accept a human name, or perhaps abstracted away to the framework level. Some form of human name formatting function should be implemented, potentially leveraging an existing library.

> I'll try to learn using github/PR asap to be more useful and provide the changes instead of only asking :)

The above suggestions potentially require architectural changes and design guidance. Tagging @smicallef

PR #320, opened 2019-07-07
Add sfp_peegeepee

bcoles:


modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+
+                if not email:
+                    continue
+
+                if not email.lower() == eventData.lower():
+                    continue
+
+                # A bit of a hack. Submit the description to sfp_names
+                # and see if it is considered to be a name.
+                name = keys.get(key)[0]
```
smicallef: This probably needs duplicate checking so that the same event for the same possible name isn't generated multiple times.

modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+                if not email:
+                    continue
+
+                mailDom = email.lower().split('@')[1]
+
+                if not self.getTarget().matches(mailDom):
+                    continue
+
+                self.sf.info('Found e-mail address: ' + email)
+                evt = SpiderFootEvent('EMAILADDR', email, self.__name__, event)
```
smicallef: In my testing this resulted in duplicates, so dup checking is needed here.

modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    # Option descriptions
+    optdescs = {
+        'timeout': 'Query timeout, in seconds.',
+        'fetch_keys': 'Retrieve PGP keys for each match.'
+    }
+
+    currentEventSrc = None
+    results = dict()
```
smicallef: Going forward please initialise to None and see comment below.

modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.__dataSource__ = 'PeeGeePee'
+        self.results = dict()
```
smicallef: Instead of using `dict()`, please use `self.tempStorage()` as this enables modules to be HX-compatible.

modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+        'timeout': 'Query timeout, in seconds.',
+        'fetch_keys': 'Retrieve PGP keys for each match.'
+    }
+
+    currentEventSrc = None
+    results = dict()
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.__dataSource__ = 'PeeGeePee'
```
smicallef: FYI this doesn't need to be set, as by default the module name will be used. It only needs to be set when the module name isn't an adequate description of the data source.

modules/sfp_peegeepee.py
```diff
@@ -0,0 +1,205 @@
...
+        'fetch_keys': False
+    }
+
+    # Option descriptions
+    optdescs = {
+        'timeout': 'Query timeout, in seconds.',
+        'fetch_keys': 'Retrieve PGP keys for each match.'
+    }
+
+    currentEventSrc = None
```
smicallef: This is only needed when multiple parts of the module submit events and need to reference the source (`sfp_arin` is an example which does that), but this module isn't doing that so it can be removed.

bcoles: Done

modules/sfp_peegeepee.py
bcoles: Removed

modules/sfp_peegeepee.py
bcoles: Fixed

modules/sfp_peegeepee.py
bcoles: Fixed

modules/sfp_peegeepee.py
bcoles: Removed duplicates using a temporary `list()` uniqued with `set()`

modules/sfp_peegeepee.py
bcoles: Removed duplicates using a temporary `list()` uniqued with `set()`

PR #339, opened 2019-07-14
Add extraction for cert SANs

bcoles:


modules/sfp_ssltools.py
```diff
@@ -97,6 +102,19 @@ def queryScan(self, domain, port):
 
         return data
 
+    # Resolve a host
```
smicallef: Please use the resolveHost code from one of the other modules, so that it's Unicode safe.

bcoles: Done

PR #343, opened 2019-07-16
Extract hostnames from SPF records

bcoles:


bcoles:
> as `RAW_DNS_RECORDS` gets produced and is processed by `sfp_dnsresolve`

Only if the `sfp_dnsresolve` module is enabled. Should modules not be self-contained, where feasible?

PR #359, opened 2019-08-26
WIP: Add bing & Google API

AmrAwadK:
- refactor: use clearer variable names and remove unnecessary conditions

### DIFF
--- a/modules/sfp_bingsearch.py
+++ b/modules/sfp_bingsearch.py
@@ -65,7 +65,7 @@ class sfp_bingsearch(SpiderFootPlugin):
             },
         )
         if results is None:
-            # Failed to talk to bing api
+            # Failed to talk to bing api or no results returned
             return None
         urls = results["urls"]
         new_links = list(set(urls) - set(self.results))
--- a/modules/sfp_bingsearch.py
+++ b/modules/sfp_bingsearch.py
@@ -9,7 +9,6 @@
 # Copyright:   (c) Steve Micallef 2013
 # Licence:     GPL
 # -------------------------------------------------------------------------------
-import requests
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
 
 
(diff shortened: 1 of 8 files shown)


AmrAwadK:
I updated the code to call the api instead of using `bingIterate`, the results before the change are:
image

The results after are:
image

Which are even cleaner : )

AmrAwadK:
Results from IP test:
image

image

I also refactored the code to contain the api call in `bingIterate`, the only remaining point is how to store the key

modules/sfp_bingsearch.py
```diff
@@ -9,24 +9,19 @@
...
-
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
 
 
 class sfp_bingsearch(SpiderFootPlugin):
-    """Bing:Footprint,Investigate,Passive:Search Engines::Some light Bing scraping to identify sub-domains and links."""
-
-
-
+    """Bing:Footprint,Investigate,Passive:Search Engines::Obtain information from bing to identify sub-domains and links."""
```
smicallef: Please add `apikey` as a flag after the `Search Engines` field

modules/sfp_bingsharedip.py
```diff
@@ -19,21 +19,22 @@
...
     }
 
     # Option descriptions
     optdescs = {
-        'cohostsamedomain': "Treat co-hosted sites on the same target domain as co-hosting?",
-        'pages': "How many pages of results to iterate through.",
-        'verify': "Verify co-hosts are valid by checking if they still resolve to the shared IP.",
-        'maxcohost': "Stop reporting co-hosted sites after this many are found, as it would likely indicate web hosting."
+        "cohostsamedomain": "Treat co-hosted sites on the same target domain as co-hosting?",
+        "pages": "Number of max bing results to request from api.",
```
smicallef: api->API

sflib.py
```diff
@@ -1299,106 +1302,45 @@ def googleIterate(self, searchString, opts=dict()):
...
+    #   "urls": a list of urls that match the query string,
+    #   "webSearchUrl": url for bing results page,
+    # }
     # Options accepted:
-    # limit: number of search result pages before returning, default is 10
-    # nopause: don't randomly pause between fetches
+    # count: number of search results to request from the api
     # useragent: User-Agent string to use
-    # timeout: Fetch timeout
+    # timeout: api call timeout
```
smicallef: api->API :)

sflib.py
```diff
@@ -1299,106 +1302,45 @@ def googleIterate(self, searchString, opts=dict()):
...
-                time.sleep(pauseSecs)
-
-            attempts = 0
-            while attempts < 3:
-                nextPage = self.fetchUrl(u'https://www.bing.com' + nextUrl,
-                                         timeout=opts['timeout'], useragent=opts['useragent'])
-                if nextPage['code'] == "400":
-                    self.error("Bing doesn't like us any more..", False)
-                    failed = True
+        response_json = json.loads(response['content'])
```
smicallef: Should be wrapped in an exception handler in case the content is not valid JSON.

sflib.py
```diff
@@ -1299,106 +1302,45 @@ def googleIterate(self, searchString, opts=dict()):
...
-                    self.refreshTorIdent()
-                    attempts += 1
-                    failed = False
-                else:
-                    break
+        if "webPages" in response_json:
+            results = {
+                "urls": [result["url"] for result in response_json["webPages"]["value"]],
+                "webSearchUrl": response_json["webPages"]["webSearchUrl"],
+            }
```
smicallef: The above two lines have assumptions that the sub-keys will exist. This will break badly if it ever changes and should be handled more cleanly.

sflib.py
```diff
@@ -1299,106 +1302,45 @@ def googleIterate(self, searchString, opts=dict()):
...
+        response = self.fetchUrl(
+            endpoint + urllib.urlencode(params),
+            timeout=opts["timeout"],
+            useragent=opts["useragent"],
+            headers={"Ocp-Apim-Subscription-Key": opts["api_key"]},
+        )
 
-        if attempts == 3:
+        if response['status'] != 'OK':
+            self.error("Failed to get a valid response from bing API", exception=False)
```
smicallef: the bing API

AmrAwadK:
Updated google search module and `googleIterate` function,

Before:
image

After:
image

AmrAwadK:
Testing `googlesearchdomain` module:
before:
image
but most of the results were inaccurate, i.e:
image

With the update we get:
image

And the results are much cleaner:
image

AmrAwadK:
pastebin working successfully
image

PR #363, opened 2019-09-01
Add sfp_tool_whatweb

bcoles:
Add `sfp_tool_whatweb` module to wrap [whatweb](https://github.com/urbanadventurer/whatweb).

Ideally this module would instead run WhatWeb over the stored HTTP headers and HTTP response body. However, Spiderfoot stores these as separate events, and WhatWeb works best when both are present.

```python
    def producedEvents(self):
        return ['RAW_RIR_DATA', 'WEBSERVER_BANNER', 'WEBSERVER_TECHNOLOGY']
```

The `WEBSERVER_BANNER` and `WEBSERVER_TECHNOLOGY` returned events are effectively a duplicate of whatever the `sfp_webserver` module would return. However, this module won't run unless the path to WhatWeb is configured, so I see no harm in adding it to the repository. Also, the output is stored as `RAW_RIR_DATA`, which is arguably more valuable, despite not producing any events.

Further efforts could be made to parse the WhatWeb output, to create additional `WEBSERVER_TECHNOLOGY` events. For example, if the `PHP` plugin matches, then PHP is likely in use. However, that's not something I'm interested in implementing any time soon. (WhatWeb has 1,500+ plugins)

### DIFF
--- a/modules/sfp_tool_whatweb.py
+++ b/modules/sfp_tool_whatweb.py
@@ -22,12 +22,14 @@ class sfp_tool_whatweb(SpiderFootPlugin):
 
     # Default options
     opts = {
+        'aggression': 1,
         'ruby_path': 'ruby',
         'whatweb_path': ''
     }
 
     # Option descriptions
     optdescs = {
+        'aggression': 'Set WhatWeb aggression level (1-4)',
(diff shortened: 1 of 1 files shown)


smicallef:
I like the idea but this feels too young to merge. Perhaps add some config options that enable some of Whatweb's features, enabling it to provide more value? Given that most users run all modules in their scans, I'm disinclined to add a tool module unless it adds value from day one.

PR #395, opened 2019-10-04
Add sfp_fsecure_riddler module

bcoles:
Add `sfp_fsecure_riddler` module.

F-Secure Riddler is apparently approaching end of life. The website has said that for a while, and doesn't appear to offer an EOL date.

The module won't run unless you specify login credentials. Registration is currently disabled. But if you have an account, this might be useful.

There's more info that could be grabbed (response data is saved as `RAW_RIR_DATA`), but I didn't put much effort into this module, due to alleged approaching EOL.

```¯\_(ツ)_/¯```

### DIFF
--- a/modules/sfp_fsecure_riddler.py
+++ b/modules/sfp_fsecure_riddler.py
@@ -18,11 +18,13 @@ class sfp_fsecure_riddler(SpiderFootPlugin):
     """F-Secure Riddler.io:Investigate,Footprint,Passive:Search Engines:apikey:Obtain network information from F-Secure Riddler.io API."""
 
     opts = {
+        'verify': True,
         'username': '',
         'password': ''
     }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/sfp_fsecure_riddler.py [hunk 1]
```diff
@@ -0,0 +1,207 @@
...
+
+            addr = result.get('addr')
+
+            if addr:
+                addrs.append(addr)
+
+            coord = result.get('cordinates')
+
+            if coord:
+                coords.append(str(coord[0]) + ', ' + str(coord[1]))
```
smicallef: I would prefer we didn't trust/assume there will always be two elements and perform a quick sanity check.

bcoles: Fixed

PR #404, opened 2019-10-08
Add analytics IDs

bcoles:
Add analytics IDs to `sfp_webanalytics`.

### DIFF
--- a/modules/sfp_webanalytics.py
+++ b/modules/sfp_webanalytics.py
@@ -319,6 +319,23 @@ class sfp_webanalytics(SpiderFootPlugin):
                                       self.__name__, parentEvent)
                 self.notifyListeners(evt)
 
+            # have-i-been-pwned Verification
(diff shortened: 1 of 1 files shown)


modules/sfp_webanalytics.py
```diff
@@ -319,6 +319,23 @@ def handleEvent(self, event):
...
+
+            # Cisco Live Domain Verification
+            # https://www.ciscolive.com/c/dam/r/ciscolive/us/docs/2016/pdf/TECCOL-2982.pdf
+            matches = re.findall(r'cisco-ci-domain-verification=([a-f0-9]+)$', eventData.strip(), re.IGNORECASE)
+            for m in matches:
+                evt = SpiderFootEvent("WEB_ANALYTICS_ID",
+                                      "Cisco Live Domain Verification: " + m,
+                                      self.__name__, parentEvent)
+                self.notifyListeners(evt)
+
```
smicallef: It just occurred to me that there is no dup handling in this module? Or is there no chance of multiple of the same IDs being found?

bcoles: Dupes are possible, but unlikely

PR #405, opened 2019-10-08
Update sfp_citadel

bcoles:
Update `sfp_citadel` style for consistency with existing modules.

### DIFF
--- a/modules/sfp_citadel.py
+++ b/modules/sfp_citadel.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 # -------------------------------------------------------------------------------
-# Name:         sfp_citadel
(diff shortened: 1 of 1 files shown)


modules/sfp_citadel.py
```diff
@@ -46,50 +47,71 @@ def watchedEvents(self):
...
-        except Exception as ex:
-            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
-            message = template.format(type(ex).__name__, ex.args)
-            self.sf.error(message, False)
 
-# End of sfp_citadel class
+        error = data.get('error')
+        message = data.get('message')
 
+        if data.get('error') == 'true':
```
smicallef: nit: I think you mean `if error == 'true'` here right?

PR #427, opened 2019-10-20
Replace `AFFILIATE_DOMAIN` with `AFFILIATE_DOMAIN_NAME`

bcoles:
* Remove `AFFILIATE_DOMAIN_UNRESOLVED` from `sfdb`
* Rename `AFFILIATE_DOMAIN` to `AFFILIATE_DOMAIN_NAME` data type in `sfdb`
* Replace old code:
  * code which reported `AFFILIATE_DOMAIN` rather then `AFFILIATE_INTERNET_NAME` (written before `AFFILIATE_INTERNET_NAME` data type existed)
  * code which reported `AFFILIATE_DOMAIN_UNRESOLVED` rather then `AFFILIATE_INTERNET_NAME_UNRESOLVED` (written before `AFFILIATE_INTERNET_NAME_UNRESOLVED` data type existed)
* Finally, replace legitimate instances of `AFFILIATE_DOMAIN` with new `AFFILIATE_DOMAIN_NAME` type


modules/sfp_dnsraw.py
```diff
@@ -144,6 +144,14 @@ def handleEvent(self, event):
                                         evt = SpiderFootEvent(evt_type, domain, self.__name__, parentEvent)
                                         self.notifyListeners(evt)
 
+                                        if self.sf.isDomain(domain, self.opts['_internettlds']):
+                                            if evt_type.startswith('AFFILIATE'):
+                                                evt = SpiderFootEvent('AFFILIATE_DOMAIN_NAME', domain, self.__name__, event)
+                                                self.notifyListeners(evt)
+                                            else:
+                                                evt = SpiderFootEvent('DOMAIN_NAME', domain, self.__name__, event)
```
smicallef: This will result in recursive behavior right? Because the module is now generating `DOMAIN_NAME`, which it's also listening for.

modules/sfp_spyonweb.py
```diff
@@ -236,7 +236,7 @@ def handleEvent(self, event):
                     self.sf.debug("Record found too old, skipping.")
                     continue
 
-                evt = SpiderFootEvent("AFFILIATE_DOMAIN", r, self.__name__, event)
+                evt = SpiderFootEvent("AFFILIATE_INETERNET_NAME", r, self.__name__, event)
```
smicallef: Typo

modules/sfp_spyonweb.py
```diff
@@ -60,8 +60,8 @@ def watchedEvents(self):
 
     # What events this module produces
     def producedEvents(self):
-        return ['CO_HOSTED_SITE', 'INTERNET_NAME', 'AFFILIATE_DOMAIN', 
-                'WEB_ANALYTICS_ID', 'DOMAIN_NAME']
+        return ['CO_HOSTED_SITE', 'INTERNET_NAME', 'AFFILIATE_INTERNET_NAME',
+                'WEB_ANALYTICS_ID', 'DOMAIN_NAME', 'AFFILIATE_DOMAIN_NAME']
```
smicallef: I can't find code generating this event?

modules/sfp_spyonweb.py
bcoles: Fixed

modules/sfp_spyonweb.py
bcoles: Fixed

modules/sfp_dnsraw.py
bcoles: Fixed

PR #431, opened 2019-10-20
Print debug message in the event of DB lock

bcoles:
This code block needs something in it.

```
# ./sf.py 
Traceback (most recent call last):
  File "./sf.py", line 64, in <module>
    from sfdb import SpiderFootDb
  File "/root/Desktop/spiderfoot/sfdb.py", line 374
    else:
       ^
IndentationError: expected an indented block
```

### DIFF
--- a/sfdb.py
+++ b/sfdb.py
@@ -370,7 +370,7 @@ class SpiderFootDb:
             if "locked" in e.args[0] or "thread" in e.args[0]:
                 # TODO: Do something smarter here to handle locked databases
                 #print("[warning] Couldn't log due to SQLite limitations. You can probably ignore this.")
-                #self.sf.fatal("Unable to log event in DB due to lock: " + e.args[0])
+                self.sf.debug("Unable to log event in DB due to lock: " + e.args[0])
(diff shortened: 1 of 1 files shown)


sfdb.py
```diff
@@ -370,7 +370,7 @@ def scanLogEvent(self, instanceId, classification, message, component=None):
             if "locked" in e.args[0] or "thread" in e.args[0]:
                 # TODO: Do something smarter here to handle locked databases
                 #print("[warning] Couldn't log due to SQLite limitations. You can probably ignore this.")
-                #self.sf.fatal("Unable to log event in DB due to lock: " + e.args[0])
+                self.sf.debug("Unable to log event in DB due to lock: " + e.args[0])
```
smicallef: Actually it needs to be print because it gets here in the first place from `self.sf.debug`.

bcoles: Is there a reasonable no-op we can use here instead in order to preserve the code format?

smicallef:
Just did the fix, thanks for flagging.

PR #449, opened 2019-11-17
sfcli - Fix syntax error

bcoles:
```
$ ./sfcli.py 
  File "./sfcli.py", line 1137
    if type(serverconfig[k]) in == str:
                                 ^
SyntaxError: invalid syntax
```


smicallef:
Fixed this in another commit as part of other fixes.

PR #451, opened 2019-11-20
Py3 - Initial tests

AmrAwadK:
Dependencies are `pytest` and `pytest-mock`, how should test requirements be added? a separate `requirements-test.txt` file?


smicallef:
Yeah, I think a separate `requirements.txt` file for tests seems reasonable, as I suspect most users won't run the tests.

PR #480, opened 2020-02-13
Black formatting

kstrauser:
This is the exact same codebase, but reformatted by Black. It should work exactly like before.

I also added a Makefile so that it's easy to re-run Black against the whole project, and told pylint to ignore some of Black's decisions that it likes to complain about.

The new .python-version file mentions Python 3.6. What do you think of that as an acceptable baseline? That gives you nice things like f-strings, which are _so_ pleasant to use!

Edit:

PS: This is against the v3.1 branch now.


smicallef:
As discussed offline, will drop this for now, prep for the 3.1 release and when that codebase is ready will look to apply black then. Thanks for the contribution though!

Issue #502, opened 2020-04-09
Whois ERROR


rad10:
Just started using this tool, and tried to scrape data from a URL that I knew a lot about, testing if it worked. It didnt get much information, but I noticed that I had more errors than scans. Along with the many ordinal errors showing up, This error kept showing up

`Unable to perform WHOIS on 011wireless.mobi: module 'whois' has no attribute 'whois'`

 So this tells me that the whois module no longer works.

rad10:
Issue was the wrong library installed. install `python-whois` rather than `whois`

PR #503, opened 2020-04-09
Fixed the only line where whois error occurs (#502)

rad10:
This change only updates the whois module to use the new functions provided by the updated whois library. I tested the webserver with this change, and I never saw any errors regarding the whois module.

### DIFF
--- a/modules/sfp_whois.py
+++ b/modules/sfp_whois.py
@@ -65,7 +65,7 @@ class sfp_whois(SpiderFootPlugin):
         try:
             data = None
             if eventName != "NETBLOCK_OWNER":
-                whoisdata = whois.whois(eventData)
+                whoisdata = whois.query(eventData)
                 if whoisdata:
                     data = whoisdata.text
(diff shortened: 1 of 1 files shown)


bcoles:
I think there's something else going on here. I wasn't able to reproduce this issue with `sfp_whois` using your input of `011wireless.mobi`.

Also, `whois.whois` is [the correct syntax](https://github.com/richardpenman/pywhois) for [python-whois](https://pypi.org/project/python-whois/).

The `whois.query` [syntax](https://github.com/richardpenman/pywhois) is used for [whois](https://pypi.org/project/whois/), which Spiderfoot does not use.

smicallef:
I've run into this myself in the past and indeed the cause was the installation of the wrong Python whois library. Are you sure you did `pip install -r requirements.txt` to get the right one?

rad10:
Ok. I see the mistake made on my part now. Sorry for the massive delay in response

PR #508, opened 2020-04-12
Add sfp_grep_app

bcoles:
Requires `sf.extractUrls()` function, currently pending merge in #507

### DIFF
--- a/modules/sfp_grep_app.py
+++ b/modules/sfp_grep_app.py
@@ -1,0 +1,214 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_grep_app
+# Purpose:     Searches grep.app API for domains, URLs and emails related to the
+#              specified domain.
+#
+# Author:      <bcoles[at]gmail[.]com>
+#
+# Created:     2020-04-12
+# Copyright:   (c) bcoles 2020
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import json
+import math
+import time
+import re
+import urllib.request, urllib.parse, urllib.error
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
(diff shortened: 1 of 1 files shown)


modules/sfp_grep_app.py
```diff
@@ -0,0 +1,210 @@
...
+        self.errorState = False
+
+        for opt in list(userOpts.keys()):
+            self.opts[opt] = userOpts[opt]
+
+    def watchedEvents(self):
+        return ["DOMAIN_NAME"]
+
+    def producedEvents(self):
+        return ["EMAILADDR", "DOMAIN_NAME", "INTERNET_NAME", "INTERNET_NAME_UNRESOLVED", "LINKED_URL_INTERNAL"]
```
smicallef, line 52: Listening for `DOMAIN_NAME` and producing it, so probably worth having a check in `handleEvent` to ensure we aren't recursive?

modules/sfp_grep_app.py
```diff
@@ -0,0 +1,210 @@
...
+        'max_pages': "Maximum number of pages of results to fetch.",
+        'dns_resolve': "DNS resolve each identified domain."
+    }
+
+    results = None
+    errorState = False
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.__dataSource__ = "grep.app"
```
smicallef, line 41: FWIW this is only needed when the data source is not the name of the module (e.g. the `sfp_spider` module overrides the data source with `Target Website` since "Web Spider" is not really meaningful as a source).

modules/sfp_grep_app.py
bcoles: The `DOMAIN_NAME` should be in `self.results` after the first run, preventing recursion.

modules/sfp_grep_app.py
bcoles: Fixed

PR #510, opened 2020-04-13
Update sfp_flickr to extract URLs and domains

bcoles:
Requires `sf.extractUrls()` function, currently pending merge in #507

### DIFF
--- a/modules/sfp_flickr.py
+++ b/modules/sfp_flickr.py
@@ -158,20 +158,20 @@ class sfp_flickr(SpiderFootPlugin):
                 return None
 
             # Calculate number of pages to retrieve
-            result_pages = int(photos.get('pages'))
+            result_pages = int(photos.get('pages', 0))
 
             if result_pages < pages:
                 pages = result_pages
 
             if 'max_allowed_pages' in photos:
-                allowed_pages = int(photos.get('max_allowed_pages'))
+                allowed_pages = int(photos.get('max_allowed_pages', 0))
                 if pages > allowed_pages:
                     pages = allowed_pages
 
             self.sf.info("Parsing page " + str(page) + " of " + str(pages))
 
             # Extract data
-            for photo in photos.get('photo'):
+            for photo in photos.get('photo', list()):
                 emails = self.sf.parseEmails(str(photo))
                 for email in emails:
                     if email in self.results:
--- a/modules/sfp_flickr.py
+++ b/modules/sfp_flickr.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 # -------------------------------------------------------------------------------
-# Name:         sfp_flickr
-# Purpose:      SpiderFoot plug-in for retrieving email addresses belonging
-#               to your target from Flickr.
+# Name:        sfp_flickr
+# Purpose:     Search Flickr API for domains, URLs and emails related to the
+#              specified domain.
 #
 # Author:      <bcoles@gmail.com>
 #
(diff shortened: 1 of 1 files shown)


modules/sfp_flickr.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 # -------------------------------------------------------------------------------
 # Name:         sfp_flickr
-# Purpose:      SpiderFoot plug-in for retrieving email addresses belonging
-#               to your target from Flickr.
+# Purpose:      Search Flickr API for domains, URLs and emails related to the
+#               specified doman.
```
smicallef, line 5: Typo

modules/sfp_flickr.py
```diff
@@ -144,23 +167,61 @@ def handleEvent(self, event):
 
             self.sf.info("Parsing page " + str(page) + " of " + str(pages))
 
-            # Extract emails
+            # Extract data
             for photo in data['photos']['photo']:
```
smicallef, line 171: I'd prefer the use of `.get()` here since there is no guarantee that `photos` or `photo` exist as keys.

modules/sfp_flickr.py
```diff
@@ -144,23 +167,61 @@ def handleEvent(self, event):
...
+                emails = self.sf.parseEmails(str(photo))
                 for email in emails:
-                    # Skip unrelated emails
-                    mailDom = email.lower().split('@')[1]
-                    if not self.getTarget().matches(mailDom, includeChildren=True, includeParents=True):
-                        self.sf.debug("Skipped address: " + email)
+                    if email in self.results:
+                        continue
+
+                    mail_domain = email.lower().split('@')[1]
```
smicallef, line 177: Worth trying to handle an exception here in case we get a malformed e-mail address?

modules/sfp_flickr.py
bcoles: `photos` is checked a few lines above.

modules/sfp_flickr.py
bcoles: `email` comes from `self.sf.parseEmails` a few lines above.

Are you suggesting the `self.sf.parseEmails` function can return malfored email addresses, and if so, what validation could be implemented here that would be superior to the validation in `self.sf.parseEmails`, and why should that validation not be applied in `self.sf.parseEmails` instead?

modules/sfp_flickr.py
bcoles: Fixed

modules/sfp_flickr.py
```diff
@@ -100,67 +130,101 @@ def handleEvent(self, event):
...
-                return None
+            photos = data.get('photos')
 
-            if 'pages' not in data['photos']:
+            if not photos:
                 return None
 
             # Calculate number of pages to retrieve
-            result_pages = int(data['photos']['pages'])
+            result_pages = int(photos.get('pages'))
```
smicallef, line 161: There's a chance this could be `None`, so the `int()` would raise an exception. Maybe set the default return value to `0` with `get()` to be on the safe side?

modules/sfp_flickr.py
```diff
@@ -100,67 +130,101 @@ def handleEvent(self, event):
...
 
             if result_pages < pages:
                 pages = result_pages
 
-            if 'max_allowed_pages' in data['photos']:
-                allowed_pages = int(data['photos']['max_allowed_pages'])
+            if 'max_allowed_pages' in photos:
+                allowed_pages = int(photos.get('max_allowed_pages'))
```
smicallef, line 167: As above, there's a chance this could be `None`, so the `int()` would raise an exception. Maybe set the default return value to `0` with `get()` to be on the safe side?

modules/sfp_flickr.py
```diff
@@ -100,67 +130,101 @@ def handleEvent(self, event):
...
                 if pages > allowed_pages:
                     pages = allowed_pages
 
             self.sf.info("Parsing page " + str(page) + " of " + str(pages))
 
-            # Extract emails
-            for photo in data['photos']['photo']:
-                emails = self.sf.parseEmails(str(bytes(photo).decode('unicode-escape')))
+            # Extract data
+            for photo in photos.get('photo'):
```
smicallef, line 174: As above, there's a chance this could be `None`, so expecting to iterate it would raise an exception. Maybe set the default return value to `list()` with `get()` to be on the safe side?

PR #517, opened 2020-04-24
Implemented sf_creditcard.py

krishnasism:
**Implemented sf_creditcard.py module**
**Changes:** 

- Developed sf_creditcard.py

- Implemented parseCreditCard(self,data) in sflib.py

- Added entry for "CREDIT_CARD_NUMBER" in sfdb.py

**Functionality:** 
Extract credit card information from the incoming "TARGET_WEB_CONTENT" data source.

### DIFF
--- a/modules/sfp_creditcard.py
+++ b/modules/sfp_creditcard.py
@@ -39,6 +39,8 @@ class sfp_creditcard(SpiderFootPlugin):
 
         # Clear / reset any other class member variables here
         # or you risk them persisting between threads.
+
+        # Override datasource for sfp_creditcard module
         self.__dataSource__ = "Target Website"
 
         for opt in list(userOpts.keys()):
--- a/modules/sfp_creditcard.py
+++ b/modules/sfp_creditcard.py
@@ -44,7 +44,6 @@ class sfp_creditcard(SpiderFootPlugin):
             self.opts[opt] = userOpts[opt]
 
     # What events is this module interested in for input
-    # For a list of all events, check sfdb.py.
     def watchedEvents(self):
         return ["TARGET_WEB_CONTENT"]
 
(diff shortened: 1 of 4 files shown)


modules/sfp_creditcard.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Clear / reset any other class member variables here
+        # or you risk them persisting between threads.
+
+        for opt in list(userOpts.keys()):
+            self.opts[opt] = userOpts[opt]
+
+    # What events is this module interested in for input
+    def watchedEvents(self):
+        return ["TARGET_WEB_CONTENT"]
```
smicallef, line 48: Do we want to limit this to only web content form the target, since it seems unlikely they'd have actual credit card numbers on their website, right? Maybe we should also look at some of the dark web or leak site content as well?

modules/sfp_creditcard.py
```diff
@@ -0,0 +1,91 @@
...
+    # Tracking results can be helpful to avoid reporting/processing duplicates
+    results = None
+
+    # Tracking the error state of the module can be useful to detect when a third party
+    # has failed and you don't wish to process any more events.
+    errorState = False
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        # Krishnasis - Note : What does self.tempStorage() return ?
```
smicallef, line 37: Think you can remove that comment now :)

modules/sfp_creditcard.py
```diff
@@ -0,0 +1,91 @@
...
+    def producedEvents(self):
+        return ["CREDIT_CARD_NUMBER"]
+
+    # Handle events sent to this module
+    def handleEvent(self, event):
+        eventName = event.eventType
+        srcModuleName = event.module
+        eventData = event.data
+
+        event.moduleDataSource="Target Website"
```
smicallef, line 60: Style nit: keep spaces between `=`, same in a few places below.

smicallef:
Nice one, thanks!

PR #519, opened 2020-04-29
Implement sfp_countryname.py

krishnasism:
**Implemented sfp_countryname.py module**
**Changes**
- Developed sfp_countryname.py module

Implemented extraction of country names from

- Phone Number 
- Domain TLDs
- IBAN Number
- WHOIS data 

Added entry 'COUNTRY_NAME' in sfdb.py

Implemented the usage of opts


modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+import phonenumbers
+from phonenumbers.phonenumberutil import region_code_for_country_code
+
+class sfp_countryname(SpiderFootPlugin):
+    """Country Name Extractor:Footprint,Investigate,Passive:Content Analysis::Identify country names in any obtained data."""
+
+    # Default options
+    opts = {
+        # options specific to this module
+        'coHosted' : True,
```
smicallef, line 26: Keep options all lower case, so that it's consistent with other modules

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+        'affiliate' : True,
+        'nonCountryTLD' : True,
+        'nonCountryTLDDefault' : "United States"
+    }
+
+    # Option descriptions
+    optdescs = {
+        'coHosted' : "Include searching of country name from CO_HOSTED_* data sources",
```
smicallef, line 34: Descriptions should not make reference to internal event IDs.. needs to be something human readable. e.g. in this case you simply want to say "Obtain country names for co-hosted sites?"

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+        'affiliate' : "Include searching of country name from AFFILIATE_* data sources",
+        'nonCountryTLD' : "Parse TLDs not associated with any country as default country domains",
+        'nonCountryTLDDefault' : "Default value for TLDs not associated with any country"
```
smicallef, line 37: The more I think about this, the more I think it doesn't make sense to do for anything beyond .com, .net, etc. since non-country specific domains covers an increasingly broad set of sites such as the new generation of domains (.app, .web, .io, etc.)

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+
+        # Clear / reset any other class member variables here
+        # or you risk them persisting between threads.
+
+        # Override datasource for sfp_countryname module
+        self.__dataSource__ = "Target Website"
```
smicallef, line 47: Probably get rid of this since you have the logic down below to override it on a case-by-case basis. There is no sane default for this module.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+
+        for opt in userOpts.keys():
+            self.opts[opt] = userOpts[opt]
+
+    # Getter method for country codes and country names dictionary
+    def getCountryCodeDict(self):
```
smicallef, line 53: Any reason why this isn't a variable rather than a function?

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+        # If any countries are found
+        if len(countries) > 0:
+            return countries
+
+        return None
+            
+    # What events is this module interested in for input
+    def watchedEvents(self):
+        return ["IBAN_NUMBER", "PHONE_NUMBER", "AFFILIATE_DOMAIN_NAME",
+                "CO_HOSTED_SITE_DOMAIN", "DOMAIN_NAME", "SIMILARDOMAIN",
```
smicallef, line 249: I think SIMILARDOMAIN is probably too far a reach from scope, since by definition such a domain will exist on a different TLD and thus probably result in a lot of false positives. At the very least this should be a configurable option, defaulted to False (you have an option for cohosts and affiliates but not similar domains).

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+            countryName = self.detectCountryFromPhone(eventData)
+        elif eventName in ["DOMAIN_NAME", "SIMILARDOMAIN"] or (eventName == "AFFILIATE_DOMAIN_NAME" and self.opts["affiliate"]) or(eventName == "CO_HOSTED_SITE_DOMAIN" and self.opts["coHosted"]):
+            countryName = self.detectCountryFromTLD(eventData)
+        elif eventName == "IBAN_NUMBER":
+            countryName = self.detectCountryFromIBAN(eventData)
+        elif eventName == "DOMAIN_WHOIS" or (eventName == "AFFILIATE_DOMAIN_WHOIS" and self.opts["affiliate"]) or (eventName == "CO_HOSTED_SITE_DOMAIN_WHOIS" and self.opts["coHosted"]):
+            countryNames = self.detectCountryFromWhoIs(eventData)
+
+        # If whois module data source is recevied, multiple country names might be there
+        if 'countryNames' in locals() and "WHOIS" in eventName:
```
smicallef, line 281: Not sure why you have `countryName` and `countryNames`.. seems like there's an opportunity to simplify there? Maybe just have `countryNames` which is a `list()` and for the functions returning single values, encapsulate the result in `[]`?

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+        srcModuleName = event.module
+        eventData = event.data
+        
+        evttype = "COUNTRY_NAME"
+
+        self.sf.debug("Received event, " + eventName + ", from " + srcModuleName)
+
+        myres = list()
+        countryName = None
+
```
smicallef, line 269: Duplication check needed here, to not process events for anything already processed.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,323 @@
...
+                if event.moduleDataSource:
+                    evt.moduleDataSource = event.moduleDataSource
+                else:
+                    evt.moduleDataSource = "Unknown"
+                self.notifyListeners(evt)
+            return None
+
+        # Note : Is checking for duplicates needed? 
+        # Since, incoming data is all unique. There is no chance of duplication
+        # Commented for now.
```
smicallef, line 300: See above comment about the duplication check for incoming events. For outgoing, I suggest to perform the check within the Whois loop, since if the same country is mentioned multiple times in a Whois record, you only want to report it once. But if different countries are mentioned, you want each one reported.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,308 @@
...
+        'noncountrytlddefault' : "United States",
+        'similardomain' : False,
+    }
+
+    # Option descriptions
+    optdescs = {
+        'cohosted' : "Obtain country name from co-hosted sites",
+        'affiliate' : "Obtain country name from affiliate sites",
+        'noncountrytld' : "Parse TLDs not associated with any country as default country domains",
+        'noncountrytlddefault' : "Default country name for TLDs not associated with any country(.com, .net)",
```
smicallef, line 38: Same for .org, .gov and .mil right?

modules/sfp_countryname.py
```diff
@@ -0,0 +1,308 @@
...
+            "AC" : "Ascension Island", "EU" : "European Union", "SU" : "Soviet Union", 
+            "UK" : "United Kingdom"
+        }
+        return abbvCountryCodes
+    
+    # Getter method for dictionary of codes not associated with any country
+    def getNonCountryCodesDict(self):
+        
+        # List of TLD not associated with any country
+        nonCountryCodes = ["COM", "NET"]
```
smicallef, line 150: As per above comment, should have a few more here.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,308 @@
...
+    def handleEvent(self, event):
+        eventName = event.eventType
+        srcModuleName = event.module
+        eventData = event.data
+        
+        evttype = "COUNTRY_NAME"
+
+        self.sf.debug("Received event, " + eventName + ", from " + srcModuleName)
+
+        myres = list()
```
smicallef, line 265: Since you are setting this to an empty list for each event, the `if` statement on line 279 below will always return `False`.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,308 @@
...
+            countryNames.remove(None)
+        if len(countryNames) == 0 :
+            return None
+
+        # Remove duplicates if there are any
+        countryNames = set(countryNames)
+
+        for countryName in countryNames:
+            self.sf.debug("Found country name : " + countryName)
+            myres.append(countryName)
```
smicallef, line 298: You are duplicating the use of myres.. in the first case you're trying to use it to avoid processing duplicate events (which is fine), but here you're also using it to avoid /producing/ duplicate events, and that will cause problems. I suggest two separate checks - the one above to avoid handling the same events, and one here which only uses a variable scoped to this function to avoid producing multiple country names from the same source event.

modules/sfp_countryname.py
```diff
@@ -0,0 +1,308 @@
...
+        # Parse source phone nummber
+        phoneNumber = phonenumbers.parse(srcPhoneNumber)
+        # Check if country code is present in the phone number
+        try:
+            countryCode = region_code_for_country_code(phoneNumber.country_code)
+        except:
+            # Region code not present in source phone number
+            self.debug("Skipped invalid phone number: " + srcPhoneNumber)
+            return None
+        return abbvCountryCodes[countryCode.upper()]
```
smicallef, line 180: If there is even a remote chance that the `countryCode` here doesn't exist in the dictionary above, best to use `abbvCountryCodes.get(countryCode.upper())`, otherwise this will raise an exception when the key doesn't exist. Using `.get()` will return None.

PR #524, opened 2020-05-04
Implement sfp_dehashed.py

krishnasism:
- Fetch JSON response from dehashed API
- Extract and send to the listener : email, password, hashed password, raw rir data


modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+
+        self.results = self.tempStorage()
+
+        # Clear / reset any other class member variables here
+        # or you risk them persisting between threads.
+
+        self.__dataSource__ = "Dehashed"
```
smicallef, line 47: By default, the module name is used as the source, so this isn't needed.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+
+        entries = list()
+        # To implement  : Number of pages 
+        # Fetch Dehashed data for incoming data (email)
+        jsonData = self.query(eventData)
+
+        if jsonData is None:
+            return None
+
+        entries = jsonData['entries']
```
smicallef, line 125: Remember to use `.get()` instead, since this will raise an exception if `entries` doesn't exist, or catch an exception instead.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+
+        if event.moduleDataSource:
```
smicallef, line 127: This clause isn't needed since the data source is always Dehashed, we don't need to inherit the data source from anywhere else.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+        email = ""
+        password = ""
+        hashed_password = ""
+        emails = list()
+        passwords = list()
+        hashed_passwords = list()
+
+        for entry in entries:
+            # If email does not exist or is null
+            if entry['email'] is None or str(entry['email']).strip() == '':
```
smicallef, line 142: Use `.get()` or catch an exception in case the `email` key doesn't exist.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+                continue
+
+            if not entry['obtained_from'] is None or str(entry['obtained_from']).strip() == '':
```
smicallef, line 145: As above.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,187 @@
...
+            if not entry['hashed_password'] is None or str(entry['hashed_password']).strip() == '':
+                hashed_password = entry['hashed_password']
+
+                if not len(password) == 0:
+                    hashed_passwords.append(email + " : " + hashed_password + "(Password : " + entry['password'] + ") [" + breachSource + "] ")
+                else:
+                    hashed_passwords.append(email + " : " + hashed_password + " [ " + breachSource + "]")
+
+            # Pass the JSON object as RAW_RIR_DATA 
+            evt = SpiderFootEvent("RAW_RIR_DATA", str(entry), self.__name__, event)
```
smicallef, line 166: You aren't calling `self.notifyListeners()` for this event.

modules/sfp_dehashed.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if jsonData is None:
+            return None
+        
+        entries = jsonData.get('entries')
+        if entries is None:
+            return None
+
+        
+        evt.moduleDataSource = event.moduleDataSource
```
smicallef, line 128: This is not correct, since the data source is Dehashed, not the source of the email address event. The line should just be removed and leave it to the default behavior

krishnasism: Okay, got it!

PR #528, opened 2020-05-11
Convert from Threading to Multiprocess to support parallel jobs

fallingcubes:
This pull request modifies several components to support multiple Spiderfoot queries to be run in parallel. This is accomplished by migrating away from  the multithreading library in python to multiprocess. By spinning off into subprocesses, the Python Global Interpreter Lock is bypassed allowing for the multiple jobs to concurrently access the database.


bcoles:
Please don't delete tests:

* test/unit/test_spiderfootscanstatus.py

If the test no longer passes then so be it. It can be updated later.

smicallef:
Thanks a lot for the contribution!

steve-bate:
Regarding multiple processes "concurrently accessing the database", just note that "SQLite allows multiple processes to have the database file open at once, and for multiple processes to read the database at once. When any process wants to write, **it must lock the entire database file for the duration of its update**." (from https://www.sqlite.org/faq.htm).

PR #529, opened 2020-05-12
Implement sfp_badpackets module

krishnasism:
Implementing malicious IP address detection & data extraction using Bad Packets API

### DIFF
--- a/modules/sfp_badpackets.py
+++ b/modules/sfp_badpackets.py
@@ -21,7 +21,6 @@ class sfp_badpackets(SpiderFootPlugin):
 
     opts = {
         'api_key': '',
-        'checkcohosts': True,
         'checkaffiliates': True,
         'subnetlookup': False,
         'netblocklookup': True,
@@ -32,7 +31,6 @@ class sfp_badpackets(SpiderFootPlugin):
     # Option descriptions. Delete any options not applicable to this module.
     optdescs = {
         "api_key": "Bad packets API Key",
-        'checkcohosts': "Check co-hosted sites?",
         'checkaffiliates': "Check affiliates?",
         'subnetlookup': "Look up all IPs on subnets which your target is a part of?",
         'netblocklookup': "Look up all IPs on netblocks deemed to be owned by your target for possible blacklisted hosts on the same target subdomain/domain?",
@@ -65,6 +63,7 @@ class sfp_badpackets(SpiderFootPlugin):
             "COUNTRY_NAME", "DESCRIPTION_CATEGORY", "DESCRIPTION_ABSTRACT"]
 
     # Check whether the IP Address is malicious using Bad Packets API
+    # https://docs.badpackets.net/#operation/query
     def queryIPAddress(self, qry, currentOffset):
         params = {
             'source_ip_address': qry.encode('raw_unicode_escape').decode("ascii", errors='replace'),
@@ -92,6 +91,7 @@ class sfp_badpackets(SpiderFootPlugin):
             self.sf.info("No Bad Packets information found")
             return None
 
+        # Error codes as mentioned in Bad Packets Documentation
         if res['code'] == '400':
             self.sf.error("Invalid IP Address", False)
             self.errorState = True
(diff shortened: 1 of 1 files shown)


modules/sfp_badpackets.py
```diff
@@ -0,0 +1,246 @@
...
+    # Parse API Response from Badpackets
+    def parseAPIResponse(self, res):
+        if res['content'] is None:
+            self.sf.info("No Bad Packets information found")
+            return None
+
+        # Error codes as mentioned in Bad Packets Documentation
+        if res['code'] == '400':
+            self.sf.error("Invalid IP Address", False)
+            self.errorState = True
```
smicallef, line 97: `errorState` should only be set when the API appears to be broken or the module is in some bad state. Basically it's a way to prevent any further querying with this module, so just a single incorrect query is not an appropriate place for it to be set.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,246 @@
...
+                    nextPageHasData = False
+                    break
+
+                count = data.get('count')
+                if count is None or int(count) == 0:
+                    nextPageHasData = False
+                    break
+
+                # Data is returned about the IP Address
+                evt = SpiderFootEvent("IP_ADDRESS", addr, self.__name__, event)
```
smicallef, line 222: You should not generate this event if the incoming event was `IP_ADDRESS` - it should only be generated when the incoming event was a `NETBLOCK_*`.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,246 @@
...
+                self.notifyListeners(evt)
+
+                results = data.get('results')
+                maliciousIPs = set()
+                for result in results:
+                    evt = SpiderFootEvent("RAW_RIR_DATA", str(result), self.__name__, event)
```
smicallef, line 228: Only one `RAW_RIR_DATA` event should be generated for the entire API response.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,246 @@
...
+                count = data.get('count')
+                if count is None or int(count) == 0:
+                    nextPageHasData = False
+                    break
+
+                # Data is returned about the IP Address
+                evt = SpiderFootEvent("IP_ADDRESS", addr, self.__name__, event)
+                self.notifyListeners(evt)
+
+                results = data.get('results')
```
smicallef, line 225: Check if this returns `None`.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,246 @@
...
+                    self.notifyListeners(evt)
+                    self.reportExtraData(event, result)
+
+                    maliciousIP = result.get('source_ip_address')
+
+                    if maliciousIP:
+                        maliciousIPs.add(maliciousIP)
+
+                for maliciousIP in maliciousIPs:
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", str(maliciousIP), self.__name__, event)
```
smicallef, line 238: Look at other modules generating `MALICIOUS_IPADDR` events - they are structured a certain way.

smicallef, line 238: If you generate an `IP_ADDRESS` event because you received a `NETBLOCK_*` event, be sure to link that as the source event (last argument to this class initialisation), and NOT the event that `handleEvent()` received.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,244 @@
...
+                                if description:
+                                    maliciousIPDesc += " - DESCRIPTION : " + str(description) + "\n"
+                            except:
+                                self.sf.debug("No description found for target")
+                            
+                            
+                            tempEvent = event
+                            # If target is a netblock_ report current IP address as target
+                            if eventName.startswith("NETBLOCK_"):
+                                tempEvent.data = str(addr)
```
smicallef, line 234: You need to use the actual event you used above for the `IP_ADDRESS` event.

modules/sfp_badpackets.py
```diff
@@ -0,0 +1,242 @@
...
+                if results is None:
+                    nextPageHasData = False
+                    break
+
+                if results:
+                    evt = SpiderFootEvent("RAW_RIR_DATA", str(results), self.__name__, event)
+                    self.notifyListeners(evt)
+                    
+                    for result in results:
+                        maliciousIP = result.get('source_ip_address')
```
smicallef, line 211: Any reason why this would be different from the IP address queried? We obviously don't want to land in a situation where the `IP_ADDRESS` event IP is different from the one stated in the `MALICIOUS_IPADDR` event.

PR #533, opened 2020-05-16
Implement sfp_snov

krishnasism:
Fetch email addresses for given target domain, from Snov.io API.


modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+from netaddr import IPNetwork
+import urllib.request, urllib.parse, urllib.error
+import json
+
+class sfp_snov(SpiderFootPlugin):
+    """Snov:Investigate,Passive:Reputation Systems:client_id,client_secret:Gather available email ids from target domain"""
+
+    opts = {
+        'client_id': '',
```
smicallef, line 23: Prefix these with api_key_ so that they get treated as credentials.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+from netaddr import IPNetwork
+import urllib.request, urllib.parse, urllib.error
+import json
+
+class sfp_snov(SpiderFootPlugin):
+    """Snov:Investigate,Passive:Reputation Systems:client_id,client_secret:Gather available email ids from target domain"""
```
smicallef, line 20: This doesn't work the way you think it does :) Replace `client_id,client_secret` with `apikey`, as that's the flag that indicates this module requires credentials.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+        
+        res = self.sf.fetchUrl(
+            'https://api.snov.io/v1/oauth/access_token?' + urllib.parse.urlencode(params),
+            headers=headers,
+            timeout=15,
+            useragent=self.opts['_useragent']
+        )
+
+        if not res['code'] == '200':
+            self.sf.error("Could not authenticate credentials", False)
```
smicallef, line 76: Indicate Snov.io in here so that the error message is more meaningful.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+            self.errorState = True 
+            return None
+        try:
+            # Extract access token from response
+            content = res.get('content')
+            if isinstance(content, str):
+                accessToken = json.loads(content).get('access_token')
+            else:
+                accessToken = res.get('access_token')
+
```
smicallef, line 86: I'm not sure what the above `if` block is supposed to be doing? Won't the response always be a string?

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+        
+        if self.errorState:
+            return None
+
+        self.sf.debug("Received event, " + eventName + ", from " + srcModuleName)
+
+        # Always check if the API key is set and complain if it isn't, then set
+        # self.errorState to avoid this being a continual complaint during the scan.
+        if self.opts['client_id'] == "" or self.opts['client_secret'] == "":
+            self.sf.error("You enabled sfp_snov but did not set an client_id and/or client_secret", False)
```
smicallef, line 137: "set an" -> "set a", and replace client_id and/or client_secret with "Client ID and/or Client Secret" so that the user can identify the configuration settings they need to change.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+        if eventData in self.results:
+            self.sf.debug("Skipping " + eventData + " as already mapped.")
+            return None
+        else:
+            self.results[eventData] = True
+
+        # Get access token from Snov IO API
+        accessToken = self.queryAccessToken()
+        if accessToken is None or accessToken == '':
+            self.sf.debug("No access token received from Snov")
```
smicallef, line 151: If this is an error it should be logged as such, and the user told something more meaningful so they know what they should do about it.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+
+        while nextPageHasData:
+            if self.checkForStop():
+                return None
+
+            data = self.queryDomainName(eventData, accessToken, currentOffset)
+            
+            if data is None:
+                self.sf.debug("No email address found for target domain")
+                return None
```
smicallef, line 166: If this is for example the third page, and we get `None`, the previous pages are discarded. Would be good to not lose the currently accumulated data.

modules/sfp_snov.py
```diff
@@ -0,0 +1,199 @@
...
+
+            if isinstance(data, str):
+                data = json.loads(data.replace("null","None").replace("'","\""))
```
smicallef, line 169: I don't know why this is needed, but if it's something weird in the Snov.io API, please comment it in the code.

modules/sfp_snov.py
krishnasism: Okay, got it. It is the one that is responsible for the API key required icon. 👍

modules/sfp_snov.py
```diff
@@ -0,0 +1,193 @@
...
+#
+# Author:      Krishnasis Mandal <krishnasis@hotmail.com>
+#
+# Created:     16/05/2020
+# Copyright:   (c) Steve Micallef
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+from netaddr import IPNetwork
```
smicallef, line 15: Not needed in this module.

modules/sfp_snov.py
```diff
@@ -0,0 +1,193 @@
...
+import urllib.request, urllib.parse, urllib.error
+import json
+
+class sfp_snov(SpiderFootPlugin):
+    """Snov:Investigate,Passive:Reputation Systems:apikey:Gather available email ids from target domain"""
```
smicallef, line 20: ids -> IDs, "from target domain" -> "from identified domains"

PR #534, opened 2020-05-19
Implement sfp_phishstats.py

krishnasism:
Determine whether an IP Address is malicious or not from PhishStats API

### DIFF
--- a/modules/sfp_phishstats.py
+++ b/modules/sfp_phishstats.py
@@ -46,7 +46,6 @@ class sfp_phishstats(SpiderFootPlugin):
         for opt in list(userOpts.keys()):
             self.opts[opt] = userOpts[opt]
         
-
     # What events is this module interested in for input
     # For a list of all events, check sfdb.py.
     def watchedEvents(self):
@@ -83,6 +82,7 @@ class sfp_phishstats(SpiderFootPlugin):
         try:
             return json.loads(res['content'])
         except:
+            self.sf.error("Ill formatted data received as JSON response", False)
             return None
          
     # Handle events sent to this module
@@ -155,7 +155,7 @@ class sfp_phishstats(SpiderFootPlugin):
                 continue
 
             if addr != maliciousIP:
-                self.sf.error("Reported address doesn't match requested, skipping")
+                self.sf.error("Reported address doesn't match requested, skipping", False)
                 continue
 
             # Data is reported about the IP Address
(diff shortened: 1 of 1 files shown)


modules/sfp_phishstats.py
```diff
@@ -0,0 +1,186 @@
...
+                maliciousIP = data[0].get('ip')
+            except:
+                # If ArrayIndex is out of bounds then data doesn't exist
+                continue
+        
+            if maliciousIP is None:
+                continue
+
+            if addr != maliciousIP:
+                self.sf.error("Reported address doesn't match requested, skipping")
```
smicallef, line 158: If you don't set `False` as the second argument, this becomes an error that aborts the scan.

modules/sfp_phishstats.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+
+        if not res['code'] == "200":
+            self.sf.debug("No information found from Phishstats for IP Address")
+            return None
+
+        try:
+            return json.loads(res['content'])
+        except:
+            return None
```
smicallef, line 86: An error should be logged if the response was broken in some way.

modules/sfp_phishstats.py
```diff
@@ -0,0 +1,186 @@
...
+
+        res = self.sf.fetchUrl(
+          'https://phishstats.info:2096/api/phishing?' + urllib.parse.urlencode(params),
+          headers=headers,
+          timeout=15,
+          useragent=self.opts['_useragent']
+        )
+
+        if not res['code'] == "200":
+            self.sf.debug("No information found from Phishstats for IP Address")
```
smicallef, line 80: Is there the possibility of another code being returned which doesn't indicate that data wasn't found, but instead indicates a different problem? e.g. `500` (server error) or `429` (throttling)?

modules/sfp_phishstats.py
```diff
@@ -0,0 +1,186 @@
...
+
+            # Data is reported about the IP Address
+            if eventName.startswith("NETBLOCK_"):
+                ipEvt = SpiderFootEvent("IP_ADDRESS", addr, self.__name__, event)
+                self.notifyListeners(ipEvt)
+
+            evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
+            self.notifyListeners(evt)
+
+
```
smicallef, line 169: nit: avoid these random line breaks throughout the module.

PR #535, opened 2020-05-20
Implement sfp_maltiverse

krishnasism:
Check if an IP Address is malicious, and gather data from Maltiverse API

### DIFF
--- a/modules/sfp_maltiverse.py
+++ b/modules/sfp_maltiverse.py
@@ -166,7 +166,7 @@ class sfp_maltiverse(SpiderFootPlugin):
             blacklistedRecords = data.get('blacklist')
 
             if blacklistedRecords is None or len(blacklistedRecords) == 0:
-                self.sf.error("No blacklist information found for IP", False)
+                self.sf.debug("No blacklist information found for IP")
                 continue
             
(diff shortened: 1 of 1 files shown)


modules/sfp_maltiverse.py
```diff
@@ -0,0 +1,218 @@
...
+          timeout=15,
+          useragent=self.opts['_useragent']
+        )
+        
+        if res['code'] == "400":
+            self.sf.error("Bad request. " + qry + " is not a valid IP Address", False)
+            return None
+
+        if res['code'] == "404":
+            self.sf.error("Not found", false)
```
smicallef, line 82: `False`, not `false`.

smicallef: Also, please be more descriptive in error messages.

modules/sfp_maltiverse.py
```diff
@@ -0,0 +1,218 @@
...
+            return None
+
+        if not res['code'] == "200":
+            self.sf.error("No information found from Maltiverse for IP Address", false)
```
smicallef, line 86: `False`, not `false`.

PR #537, opened 2020-05-25
Implement sfp_keybase

krishnasism:


modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+          useragent=self.opts['_useragent']
+        )
+
+        # In this case, it will always be 200 if keybase is queried 
+        # The actual response codes are stored in status tag of the response
+        if not res['code'] == '200':
+            return None
+        
+        # Replacing null with "None"
+        content = json.loads(str(res['content']).replace("null", "\"None\""))
```
smicallef, line 71: This has the chance to break things, e.g. imagine a username with null in it, or other valid places where null might be part of a bigger string. Can you make the pattern matching more specific to reduce that risk?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+
+        status = content.get('status')
+
+        code = status.get('code')
```
smicallef, line 75: Check that `status` isn't `None` before doing this.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+            self.sf.debug("Skipping " + eventData + " as already mapped.")
+            return None
+
+        self.results[eventData] = True
+
+        userName = eventData
+
+        # Extract username if a Keybase link is received 
+        if eventName == "LINKED_URL_EXTERNAL":
+            linkRegex = "keybase.io\/[A-Za-z0-9]+"
```
smicallef, line 104: Can usernames not contain any other characters? e.g. `-` or `_`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+            evt = SpiderFootEvent("USERNAME", str(userName), self.__name__, event)
+            self.notifyListeners(evt)    
+    
+        
+        # Replacing string values that are not enclosed within double quotes
+        # Also replacing values like True and False to their corresponding numeric values
+        # If the above steps aren't performed, json.loads() fails
+
+        # Contains all data about username
+        them = json.loads(str(content.get('them')[0]).replace("'", "\"").replace("True", "1").replace("False", "0"))
```
smicallef, line 136: the `.get('them')[0]` here assumes the result of the `get()` is a list, which might not be the case for whatever reason. Best to pass a second argument to `get()` as `list()`, which will make `get()` return that instead of `None`.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+        
+        if them is None or them == "None":
+            self.sf.debug("No data found for username")
+            return None
+
+        # Basic information about the username
+        basics = json.loads(str(them.get('basics')).replace("'", "\""))
```
smicallef, line 143: This and the following line should be in a `try` block to capture any JSON parsing errors.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+        if basics:
+            responseUserName = basics.get('username')
+            if not userName == responseUserName:
+                self.sf.error("Username does not match received response, skipping", False)
+                return None
+        
+        if profile:
+            # Get and report full name of user
+            fullName = profile.get('full_name')
+            if not fullName == "None":
```
smicallef, line 158: Why `"None"` and not `None`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+                evt = SpiderFootEvent("HUMAN_NAME", str(fullName), self.__name__, event)
+                self.notifyListeners(evt)
+            
+            # Get and report location of user
+            location = profile.get('location')
+            if not location == "None":
```
smicallef, line 164: Why `"None"` and not `None`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+                evt = SpiderFootEvent("GEOINFO", str(location), self.__name__, event)
+                self.notifyListeners(evt)
+            
+            # Extract social media information from JSON response
+            socialMediaLinksRegex = ["https:\/\/github.com\/[A-Za-z0-9-_.]+", "https:\/\/twitter.com\/[A-Za-z0-9-_.]+",
```
smicallef, line 169: Perhaps use the same ones from `sfp_social`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+                if len(links) == 0:
+                    continue
+                
+                for link in links:
+
+                    if link in self.results:
+                        continue
+                    self.results[link] = True
+
+                    evt = SpiderFootEvent("SOCIAL_MEDIA", str(link), self.__name__, event)
```
smicallef, line 184: Look at the format of the `SOCIAL_MEDIA` event from `sfp_social`, as it needs to be the same.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,203 @@
...
+                    self.notifyListeners(evt)
+
+        # Get cryptocurrency addresses 
+        cryptoAddresses = json.loads(str(them.get('cryptocurrency_addresses')).replace("'", "\""))
+        
+        # Extract and report bitcoin addresses if any
+        if cryptoAddresses:
+            bitcoinAddresses = json.loads(str(cryptoAddresses.get('bitcoin')).replace("'", "\""))
```
smicallef, line 192: Any other crypto addresses available? e.g. Ethereum?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+          useragent=self.opts['_useragent']
+        )
+
+        # In this case, it will always be 200 if keybase is queried 
+        # The actual response codes are stored in status tag of the response
+        if not res['code'] == '200':
+            return None
+        
+        # Replacing null with "None"
+        content = json.loads(str(res['content']).replace(":null", ":\"None\""))
```
smicallef, line 72: why the `str()` and then replacing null with None? You should be able to just `json.loads(res['content'])`

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+        # Extract username if a Keybase link is received 
+        if eventName == "LINKED_URL_EXTERNAL":
+            linkRegex = "keybase.io\/[A-Za-z0-9-_.]+"  
+            link = re.findall(linkRegex, eventData)
+
+            if len(link) == 0:
+                self.sf.debug("Not a keybase link")
+                return None
+
+            userName = link.split("/")[1]
```
smicallef, line 120: The link will likely be https://keybase.io/username, possibly even https://keybase.io/username/something, so shouldn't this handle that? I suggest using a regex to pull out the username, regardless of how it's prefixed or suffixed.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+
+        if content is None:
+            self.sf.debug("No data found for username")
+            return None
+        
+        evt = SpiderFootEvent("RAW_RIR_DATA", str(content), self.__name__, event)
+        self.notifyListeners(evt) 
+
+        if eventName == "LINKED_URL_EXTERNAL":
+            evt = SpiderFootEvent("USERNAME", str(userName), self.__name__, event)
```
smicallef, line 136: Why `str()`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+        
+        # Replacing string values that are not enclosed within double quotes
+        # Also replacing values like True and False to their corresponding numeric values
+        # If the above steps aren't performed, json.loads() fails
+        
+        # Checking with == "None" in addition to is None, because null values are replaced with "None"
+
+        # Contains all data about the target username
+        try:
+            them = json.loads(str(content.get('them')[0]).replace("'", "\"").replace("True", "1").replace("False", "0"))
```
smicallef, line 148: I suspect the `str()` and subsequent `replace()`s are unnecessary since `content.get('them')[0]` is going to be valid JSON (if not it should raise an exception anyway).

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+        except:
+            them = None
+
+        if them is None or them == "None":
```
smicallef, line 152: Checking for "None" as a string should not be necessary.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+            self.sf.debug("No data found for username")
+            return None
+
+        # Basic information about the username
+        try:
+            basics = json.loads(str(them.get('basics')).replace("'", "\""))
```
smicallef, line 158: Not sure why you need to replace the quote style on a valid JSON string produced by `json.loads()`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+        except:
+            basics = None
+
+        # Profile information about the username
+        try:
+            profile = json.loads(str(them.get('profile')).replace("'", "\""))
```
smicallef, line 164: As above.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+        if basics:
+            responseUserName = basics.get('username')
+            if not userName == responseUserName:
+                self.sf.error("Username does not match received response, skipping", False)
+                return None
+        
+        if profile:
+            # Get and report full name of user
+            fullName = profile.get('full_name')
+            if not (fullName is None or fullName == "None"):
```
smicallef, line 178: Should not check for "None" as a string.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+                evt = SpiderFootEvent("HUMAN_NAME", str(fullName), self.__name__, event)
```
smicallef, line 179: Why the `str()`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+                self.notifyListeners(evt)
+            
+            # Get and report location of user
+            location = profile.get('location')
+            if not (location is None or location == "None"):
+                evt = SpiderFootEvent("GEOINFO", str(location), self.__name__, event)
```
smicallef, line 185: Why the `str()`?

modules/sfp_keybase.py
```diff
@@ -0,0 +1,244 @@
...
+                self.notifyListeners(evt)
+            
+            # Extract social media information from JSON response           
+            socialMediaRegexDict = {
+                "Github": "https:\/\/github.com\/[A-Za-z0-9-_.]+",
+                "Twitter": "https:\/\/twitter.com\/[A-Za-z0-9-_.]+",
+                "Facebook": "https:\/\/facebook.com\/[A-Za-z0-9-_.]+"
+            }
```
smicallef, line 193: Please add all the other regexps as per `sfp_social`

modules/sfp_keybase.py
krishnasism: The regex will take care of that. It will extract keybase.io/username from any string.

modules/sfp_keybase.py
```diff
@@ -0,0 +1,241 @@
...
+            self.sf.debug("Skipping " + eventData + " as already mapped.")
+            return None
+
+        self.results[eventData] = True
+
+        userName = eventData
+
+        # Extract username if a Keybase link is received 
+        if eventName == "LINKED_URL_EXTERNAL":
+            linkRegex = "keybase.io\/[A-Za-z0-9\-_\.]+"
```
smicallef, line 113: May as well include `https?://` before `keybase.io` so you don't pick up external links with keybase.io, e.g. `https://someotherdomain.com/somepage.html?somevar=keybase.io/abc123`

PR #543, opened 2020-06-09
Added locks to prevent multithread access to db

steve-bate:
When running spiderfoot scans, I noticed issues with not being able to write log records and not being able to delete a stored scan while another one was running. This issue was aggravated by modules that spawn many threads (like sfp_accounts). Since sqlite doesn't allow multithreaded access, I've added a lock around the database-related code to prevent multiple threads from using the sqlite cursor at the same time. In a couple of places, I moved the commit invocation into the try/catch and protected section of code. (Note that most of the additions and deletions reported in the GitHub diff are whitespace differences. The actual changes are not that numerous.)

### DIFF
--- a/sfdb.py
+++ b/sfdb.py
@@ -13,6 +13,7 @@
 import sqlite3
 import re
 import time
+import threading
 from sflib import SpiderFoot
 
 # SQLite doesn't support regex queries, so we create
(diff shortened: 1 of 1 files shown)


smicallef:
Nice one, thanks!

PR #547, opened 2020-06-12
Changed account existence checks. Site check performance improvements.

steve-bate:
Some of the sites in the WebBreacher site list return a 200 status code for both a found account and a missing account. The previous code was considering the account to be found if the status code matched without further testing the account_existence_string. This was resulting in false positives. I simplified the check so that if the status code and the account_existence_string match, then the account is considered found. Otherwise, the account wasn't found. It is possible that neither the account found and account missing conditions are satisfied, but since the check is not using 3-state logic, I defaulted to either found or not found.

For the site scanning, the previous code was batching groups of sites and running the batches sequentially. This meant that a slow site would hold up the current batch and delay subsequent batches. Also, each batch starts with _maxThreads and ends with zero threads. Towards the end of the batch, fewer and fewer threads are scanning sites.

I changed the design to use a single thread-safe queue and a set of worker threads (up to _maxThreads). The sites are loaded into the queue and the worker threads continue to pull sites and scan them as fast they can until the queue is empty. This way all _maxThreads threads are fully occupied throughout the entire site scan.

There are other optimizations that could be done, but with this relatively simple change, I was seeing a 2-3x speedup compared to the batched design. I also added a simple log message for monitoring the scan speed that we can use for evaluating future improvements.

### DIFF
--- a/modules/sfp_accounts.py
+++ b/modules/sfp_accounts.py
@@ -13,6 +13,7 @@
 
 import time
 import threading
+from queue import Queue, Empty as QueueEmpty
 import json
 import random
 from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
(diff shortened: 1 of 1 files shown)


smicallef:
Nice, thanks!

PR #551, opened 2020-06-14
Fixed webui query route. Used redirects instead of rerender for some routes.

steve-bate:
The query route was having problems with utf-8 encoding of the results and was not able to JSON-encode a list with tuples. I changed the result to a list of dictionaries and enabled utf-8 encoding in CherryPy.

For the webui routes that invoke an action and then show a page for monitoring (rescan, etc.), I changed the behavior to invoke the action and then redirect to the monitoring page rather than rendering the target page in the handler. This helps to prevent accidental resubmission if the user refreshes the page and makes the monitoring page URL bookmark-friendly. 

I also corrected the pageid value for some of the template rendering functions.

### DIFF
--- a/sf.py
+++ b/sf.py
@@ -383,10 +383,16 @@ if __name__ == '__main__':
 
     # Enable access to static files via the web directory
     currentDir = os.path.abspath(sf.myPath())
-    conf = {'/static': {
-        'tools.staticdir.on': True,
-        'tools.staticdir.dir': os.path.join(currentDir, 'static')
-    }}
+    conf = {
+        '/query': {
+            'tools.encode.text_only': False,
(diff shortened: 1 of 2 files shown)


steve-bate:
I'll squash the commits.

smicallef:
What's the rationale behind the `pageid` change? It's used by `dyn/HEADER.tmpl` to know what section its in, and therefore what section to mark as active in the header.

steve-bate:
The weak rationale (based on limited understanding) was to make the pageid consistent with the template being rendered. I will revert those changes until I understand the pageid usage better.

steve-bate:
Updated

steve-bate:
Any other questions or comments on this?

Issue #514, opened 2020-04-20
[Suggestion] Add security.txt parsing to sfp_spider module


bcoles:
Modify the `sfp_spider` module to parse `/.well-known/security.txt` and store the associated email address and optionally also store the PGP key.

PR #552, opened 2020-06-14
A module that will retrieve and parse security.txt files.

steve-bate:
#514

### DIFF
--- a/modules/sfp_securitytxt.py
+++ b/modules/sfp_securitytxt.py
@@ -1,0 +1,108 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_securitytxt
+# Purpose:     Find and parse .well-known/security.txt files.
+#
+# Author:      Steve Bate <svc-spiderfoot@stevebate.net>
+#
+# Created:     14/06/2020
+# Copyright:   (c) Steve Micallef
+# Licence:     GPL
+#
+# See also:    https://tools.ietf.org/html/draft-foudil-securitytxt-09#page-7
+# -------------------------------------------------------------------------------
+
+from sflib import SpiderFootPlugin, SpiderFootEvent
+from urllib.parse import urlparse, urlunparse
+from securitytxt import SecurityTxt
+
+
+class sfp_securitytxt(SpiderFootPlugin):
+    """security.txt:Footprint:Crawling and Scanning::Retrieves and parses /.well-known/security.txt files"""
+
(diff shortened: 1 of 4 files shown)


steve-bate:
The text fixtures/stubs that I used to unit test the module could be pulled into a test-support library (and extended) to support testing other modules.

modules/sfp_securitytxt.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_securitytxt
+# Purpose:     Find and parse .well-known/security.txt files.
+#
+# Author:      Steve Micallef <steve@binarypool.com>
```
smicallef, line 6: This should reflect you as the author :)

smicallef:
Style-wise there are many differences to other modules. Could you refactor the module to fit with the others? I know the style is not ideal, but I'd rather they all be consistent and then think of an ideal new structure as a separate effort.

modules/sfp_securitytxt.py
steve-bate: I originally had my name there, but I wasn't sure how to handle the copyright. Should I just replace the name and leave you as the copyright holder?

steve-bate:
Can you give me a few examples about stylistic differences that would concern you?

modules/sfp_securitytxt.py
smicallef: Your name there as the author is fine.

smicallef:
Take a look at other modules, specifically `setup()` and note there is no `__init__()` in the other modules. Also naming like `self.sf` instead of `self.spiderfoot` and others.

steve-bate:
OK, thanks for the clarification. I added the __init__ because of some static analysis warnings (about creating instance fields outside of __init__), but I can remove it and disable the analysis warnings. I didn't realize the spiderfoot vs sf naming would be an issue since the field is private to this module and I prefer to use expressive variable names. I'm guessing that you maybe didn't like the names of some of the other private fields and function parameters too.

smicallef:
Keep in mind that SpiderFoot has evolved from around 2012 when I didn't know Python at all, so there are a lot of relics in the code that could be re-written. At this stage however I'd prefer that the modules are consistent so that eventually it will be a simpler exercise to convert them to a new structure en masse.

steve-bate:
I understand. I'll pull the PR and resubmit it after I rewrite it.

PR #553, opened 2020-06-16
Implement sfp_spur & sfp_twilio

krishnasism:


modules/sfp_spur.py
```diff
@@ -0,0 +1,222 @@
...
+                if city:
+                    geoInfo += city + ", "
+                
+                if state:
+                    geoInfo += state + ", "
+                
+                if country:
+                    geoInfo += country
+                
+                evt = SpiderFootEvent("GEOINFO", geoInfo, self.__name__, event)
```
smicallef, line 183: Should probably use `ipEvt` instead of `event`, since it was the IP address you looked up, not the netblock.

modules/sfp_spur.py
```diff
@@ -0,0 +1,222 @@
...
+                self.notifyListeners(evt)
+            
+            asData = data.get('as')
+
+            if asData:
+                orgName = asData.get('organization')  
+
+                if orgName:
+                    evt = SpiderFootEvent("COMPANY_NAME", orgName, self.__name__, event)
```
smicallef, line 192: Should probably use `ipEvt` instead of `event`, since it was the IP address you looked up, not the netblock.

modules/sfp_spur.py
```diff
@@ -0,0 +1,222 @@
...
+            data = json.loads(self.queryIPAddress(addr))
+
+            if data is None:
+                continue
+            
+            if eventName.startswith("NETBLOCK_"):
+                ipEvt = SpiderFootEvent("IP_ADDRESS", addr, self.__name__, event)
+                self.notifyListeners(evt)
+
+            evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef, line 163: Should probably use `ipEvt` instead of `event`, since it was the IP address you looked up, not the netblock.

modules/sfp_twilio.py
```diff
@@ -0,0 +1,126 @@
...
+
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+import base64
+import json
+
+class sfp_twilio(SpiderFootPlugin):
+    """Twilio:Footprint,Investigate,Passive:Search Engines:apikey:Obtain data from Twilio from target phone number."""
+    
+    opts = {
+        'account_sid': '',
```
smicallef, line 21: Prefix these with `api_key_` so that they are treated as credentials.

modules/sfp_twilio.py
```diff
@@ -0,0 +1,126 @@
...
+# Copyright:   (c) Steve Micallef
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+import base64
+import json
+
+class sfp_twilio(SpiderFootPlugin):
+    """Twilio:Footprint,Investigate,Passive:Search Engines:apikey:Obtain data from Twilio from target phone number."""
```
smicallef, line 18: "Obtain information from Twilio about phone numbers."

modules/sfp_twilio.py
```diff
@@ -0,0 +1,126 @@
...
+        
+        if res['code'] == '429':
+            self.sf.error("API usage limit reached.", False)
+            return None
+
+        if res['code'] == '503':
+            self.sf.error("Service unavailable.", False)
+            return None
+
+        if not res['code'] == '200':
```
smicallef, line 77: If the phone number is not found, what's the return code? If 404 then this shouldn't be an error.

modules/sfp_twilio.py
```diff
@@ -0,0 +1,126 @@
...
+
+        for opt in list(userOpts.keys()):
+            self.opts[opt] = userOpts[opt]
+
+    def watchedEvents(self):
+        return ["PHONE_NUMBER"]
+
+    # What events this module produces
+    def producedEvents(self):
+        return ["HUMAN_NAME", "COMPANY_NAME", "RAW_RIR_DATA"]
```
smicallef, line 45: Module is only reporting HUMAN_NAME and RAW_RIR_DATA, so you can drop COMPANY_NAME.

modules/sfp_twilio.py
```diff
@@ -0,0 +1,126 @@
...
+        
+        data = json.loads(content)
+
+        evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
+        self.notifyListeners(evt)
+        
+        callerName = data.get('caller_name')
+        
+        if callerName:
+            evt = SpiderFootEvent("HUMAN_NAME", callerName, self.__name__, event)
```
smicallef, line 121: Will this always be a name or sometimes a company name? Does the API distinguish between the two?

modules/sfp_spur.py
```diff
@@ -0,0 +1,237 @@
...
+    # What events is this module interested in for input
+    # For a list of all events, check sfdb.py.
+    def watchedEvents(self):
+        return ["IP_ADDRESS", "NETBLOCK_OWNER", "NETBLOCK_MEMBER",
+            "AFFILIATE_IPADDR"]
+
+    # What events this module produces
+    def producedEvents(self):
+        return ["IP_ADDRESS", "MALICIOUS_IPADDR", "RAW_RIR_DATA",
+            "GEO_INFO", "COMPANY_NAME", "AFFILIATE_MALICIOUS_IPADDR"]
```
smicallef, line 59: `MALICIOUS_AFFILIATE_IPADDR` - you can check `sfdb.py` in future to ensure the correct names are used.

modules/sfp_spur.py
```diff
@@ -0,0 +1,237 @@
...
+
+        code = res.get('code')
+
+        if code == '403':
+            self.sf.error("Invalid credentials. Please check API Token", False)
+            self.errorState = True
+            return None
+
+        if not code == '200':
+            self.sf.error("Unable to fetch data from spur.us", False)
```
smicallef, line 85: If it's a 404, we don't want an error reported.

modules/sfp_spur.py
```diff
@@ -0,0 +1,237 @@
...
+                if city:
+                    geoInfo += city + ", "
+                
+                if state:
+                    geoInfo += state + ", "
+                
+                if country:
+                    geoInfo += country
+                
+                if eventName.startswith("NETBLOCK_"):
```
smicallef, line 187: I don't think we should be reporting GEOINFO for affiliates.

modules/sfp_spur.py
```diff
@@ -0,0 +1,237 @@
...
+
+            if asData:
+                orgName = asData.get('organization')  
+
+                if orgName:
+                    if eventName.startswith("NETBLOCK_"):
+                        evt = SpiderFootEvent("COMPANY_NAME", orgName, self.__name__, ipEvt)
+                        self.notifyListeners(evt)
+                    else:
+                        evt = SpiderFootEvent("COMPANY_NAME", orgName, self.__name__, event)
```
smicallef, line 204: Likewise, for affiliates, companies shouldn't be reported.

modules/sfp_spur.py
```diff
@@ -0,0 +1,237 @@
...
+                    if operatorName:
+                        maliciousIPDesc += operatorName + ", "
+
+                maliciousIPDesc = maliciousIPDesc.strip(", ")
+
+                if eventName.startswith("NETBLOCK_"):
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", maliciousIPDesc, self.__name__, ipEvt)
+                    self.notifyListeners(evt)
+                elif eventName.startswith("AFFILIATE_"):
+                    evt = SpiderFootEvent("AFFILIATE_MALICIOUS_IPADDR", maliciousIPDesc, self.__name__, event)
```
smicallef, line 229: As per earlier comment, this needs to be corrected.

modules/sfp_twilio.py
```diff
@@ -0,0 +1,130 @@
...
+          timeout=15,
+          useragent=self.opts['_useragent']
+        )
+
+        if res['code'] == '400':
+            self.sf.error("Bad request.", False)
+            return None
+        
+        if res['code'] == '404':
+            self.sf.error("Invalid phone number.", False)
```
smicallef, line 70: Would 404 really be an invalid phone number? Wouldn't it just be one that wasn't found? In that case, it shouldn't be an error, just a debug message.

modules/sfp_spur.py
```diff
@@ -0,0 +1,243 @@
...
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", maliciousIPDesc, self.__name__, ipEvt)
+                    self.notifyListeners(evt)
+                elif eventName.startswith("AFFILIATE_"):
+                    evt = SpiderFootEvent("MALICIOUS_AFFILIATE_IPADDR", maliciousIPDesc, self.__name__, event)
+                    self.notifyListeners(evt)
+                else:
+                    evt = SpiderFootEvent("MALICIOUS_IPADDR", maliciousIPDesc, self.__name__, event)
+                    self.notifyListeners(evt)
+
+            return None
```
smicallef, line 241: From what I can see this is too indented, so it will return before processing the next iteration of the `for` loop above.

Issue #478, opened 2020-02-03
Move config files and project files to subdirectory to save them using docker volume


sidahmed-malaoui:
If you move config files and project files to a sub directory (let's call it for example `data`), it would be easier to save application state by using docker volumes this way :
```bash
docker volume create spiderfoot-vol
docker run -v spiderfoot-vol:/home/spiderfoot/data -d spiderfoot
```
This time, we can save the application state (configuration and projects) each time we launch the docker container.

PS : Your tool rocks, I'm really thankful for this wonderful work.

smicallef:
Have you tried using the latest Dockerfile for meeting this requirement?

sidahmed-malaoui:
Yes, and it's working now @smicallef . Thank you for implementing this dude

smicallef:
Credit goes to @steve-bate!

PR #558, opened 2020-06-17
Added support for configurable SpiderFoot data directory.

steve-bate:
Added a dataPath method that queries a SPIDERFOOT_DATA environment variable for an optional override to where data files are stored. The default behavior is the same as before where all data and configs are stored in the sflib.myPath location. This change will allow Docker-based deployments to specify a Docker volume for data/config storage, allowing non-destructive container upgrades. #478


smicallef:
Nice!

PR #562, opened 2020-06-18
Document sflib functions

bcoles:
This adds documentation for the majority of sflib. Some items are marked `TBD` for now.

Also ninja adds a couple of guard clauses.


smicallef:
Nice!!

PR #574, opened 2020-06-19
Dockerfile: support volumes

steve-bate:


steve-bate:
This Dockerfile adds the ability to use Docker volumes since the data is stored outside the installation directory. It's also intended to be the basis for further discussion. I know the existing Dockerfile uses a zip file of a specific version to build the image. I wasn't clear why that was done that way. This Dockerfile creates the image based on the git workspace where it is run. It also changes the ENTRYPOINT to include the sf.py script to make it easier to use temporary containers from the command line. However, I'm fine with moving it back to CMD if you prefer. With some shell aliases, that could work just as well from the command line.

steve-bate:
I've pushed an image to https://hub.docker.com/repository/registry-1.docker.io/stephenbate/spiderfoot, if you want to try one without building it yourself.

Dockerfile
```diff
@@ -6,15 +6,34 @@
...
+#
+#   sudo docker run -p 5001:5001 -v /mydir/spiderfoot:/var/lib/spiderfoot spiderfoot
+#
+# Running spiderfoot commands without web server (can optionally specify volume)
+#
+#   sudo docker run --rm spiderfoot -h
+#
+# Running spiderfoot unit tests in container
+#   
+#   sudo docker run --rm --entrypoint='/usr/bin/python3' spiderfoot -m unittest discover -s test/unit
```
bcoles, line 27:
```suggestion
#   sudo docker run --rm --entrypoint='/usr/bin/env python3' spiderfoot -m unittest discover -s test/unit
```

maybe ?

steve-bate: Can you elaborate on your rationale? My understanding is that this would be used in shebang lines to allow a user's PATH to specify different python3 interpreter than the system default. In this case, there's only one interpreter in the image. Also, the --entrypoint CLI arg doesn't allow passing arguments. We'd need to do something like:

`docker run --rm --entrypoint=/usr/bin/env spiderfoot python3 -m unittest discover -s test/unit`

If we leave the ENTRYPOINT directive in the Dockerfile as ["/usr/bin/python3"] (vs. my change), the advantage is this would become:

`sudo docker run --rm spiderfoot -m unittest discover -s test/unit`

at the expense of needing to specify the sf.py file for regular command line operation. For example,

`sudo docker run --rm spiderfoot sf.py -M`

to list modules. The image basically becomes a Python interpreter customized for spiderfoot.

bcoles: Fair enough

steve-bate:
Here's another example I should add...

`docker run -it --rm --entrypoint=python3 spiderfoot sfcli.py -s MYHOST:5001`

The discussion is causing me to reconsider adding sf.py to the ENTRYPOINT directive. I'm going to move it back to the CMD and update the examples. That would make this example be a little simpler:

`docker run -it --rm spiderfoot sfcli.py -s MYHOST:5001`

Dockerfile
```diff
@@ -26,21 +49,19 @@ RUN apk --update add --no-cache --virtual build-dependencies gcc git curl py3-pi
     && addgroup spiderfoot \
     && adduser -G spiderfoot -h /home/spiderfoot -s /sbin/nologin \
                -g "SpiderFoot User" -D spiderfoot \
-    && rmdir /home/spiderfoot \
-    && cd /home \
-    && curl -sSL https://github.com/smicallef/spiderfoot/archive/v3.0.tar.gz \
```
smicallef, line 31: Was this intentionally removed? Where will the actual SpiderFoot installation be if it's not downloaded here?

steve-bate: It's installing the files from the git workspace (see the "Copy . . " directive). This is a typical Docker workflow. If someone wants a v3.0 Docker image they checkout the v3.0 tag and build it. If they want an image with the latest changes, they checkout master and build that. For example, my spiderfoot fork is automatically building each commit using github actions and pushing them to Docker Hub with a tag based on the commit timestamp and git SHA.

smicallef: Makes sense, thanks!

PR #586, opened 2020-06-21
Add sfp_jsonwhois

bcoles:


modules/sfp_jsonwhois.py
```diff
@@ -0,0 +1,248 @@
...
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import json
+import time
+import urllib.request, urllib.parse, urllib.error
+from sflib import SpiderFoot, SpiderFootPlugin, SpiderFootEvent
+
+class sfp_jsonwhois(SpiderFootPlugin):
+    """JsonWHOIS:Footprint,Investigate,Passive:Search Engines:apikey:Search JsonWHOIS for WHOIS records associated with a domain."""
```
smicallef, line 19: Perhaps rename to JsonWHOIS.com? There is a JsonWHOIS.io too, so may confuse users when obtaining the API key.

modules/sfp_jsonwhois.py
```diff
@@ -0,0 +1,248 @@
...
+
+        self.sf.debug("Received event, %s, from %s" % (eventName, srcModuleName))
+
+        res = self.queryDomain(eventData)
+
+        if res is None:
+            self.sf.debug("No information found for domain %s" % eventData)
+            return None
+
+        #evt = SpiderFootEvent('RAW_RIR_DATA', str(res), self.__name__, event)
```
smicallef, line 147: Any particular reason you don't want to submit this event?

bcoles: It doesn't make much difference. I'll change it.

I chose not to store it as it is large and mostly duplicated. The JSON data contains the raw WHOIS record. The remainder of the JSON data is almost entirety the same data, but parsed into keys. A lot of this data is also stored as events.

Also, other modules parse `RAW_RIR_DATA`, which then results in quadrupled data.

modules/sfp_jsonwhois.py
bcoles: Done

PR #590, opened 2020-06-22
tests: re-add test/unit/test_spiderfootscanstatus.py

bcoles:
So long as the `SpiderFootScanStatus` class exists in `sflib` there should be tests for it. Please stop deleting the tests. #566 

Rather than delete these tests, they should be improved, perhaps by using the global `globalScanStatus` variable.


smicallef:
My bad, this was submitted too early but you'll see the latest changes void the need for these tests now.

PR #605, opened 2020-06-24
sfdb: Ensure search function returns a list

bcoles:
Fix incomplete patch in #604.

### DIFF
--- a/sfdb.py
+++ b/sfdb.py
@@ -327,8 +327,11 @@ class SpiderFootDb:
             list: search results
         """
 
+        if not isinstance(criteria, dict):
+            return list()
+
         if list(criteria.values()).count(None) == 3:
-            return []
+            return list()
 
         qvars = list()
         qry = "SELECT ROUND(c.generated) AS generated, c.data, \


sfdb.py
```diff
@@ -327,8 +327,11 @@ def search(self, criteria, filterFp=False):
             list: search results
         """
 
+        if not isinstance(criteria, dict):
+            return list()
```
steve-bate, line 331: If the goal is to return a list for any error, maybe the body of the method should be wrapped in a try/except where any exception returns an empty list? Currently, the code will return None for a database error, which could be caused by incorrect keys in the dictionary. The dict type-checking also defeats duck-typing (although it's unlikely that would be an issue here).

sfdb.py
```diff
@@ -327,8 +327,11 @@ def search(self, criteria, filterFp=False):
...
+
         if list(criteria.values()).count(None) == 3:
```
steve-bate, line 333: This check seems to make some assumptions about the content of the criteria dictionary (key count, valid keys, etc.). Separate arguments/keywords for the search criteria would be safer.

bcoles: This check is existing code outside of this PR. Yes, it could be approved.

sfdb.py
bcoles: > If the goal is to return a list for any error

The goal of #604 appeared to be to ensure that the `search` function returned a `list` regardless of invalid input, but did not account for an invalid input type. This PR fixes that oversight. Yes, there's still more room for improvement of error handling for the dict keys in this function.

> maybe the body of the method should be wrapped in a try/except where any exception returns an empty list
> Currently, the code will return None for a database error

It's on the TODO list. You have likely noticed that a lot of the recent commits relate to error handling and associated unit tests.

This pattern is systemic throughout `sfdb`. Currently none of the functions (outside of `init`) will `raise` (intentionally).

There is a lot of room for improvement in this library.

The database library functions should be bulletproof. The implementation should be consistent. This probably means implementing `raise` in every function and ensuring the callers use `try` as you suggest, but I would prefer more granular error handling than just shoving the entire contents into a `try` / `except`.

steve-bate: That's fine and I understand. My goal with #640 was just to fix an issue with the existing check that I ran across while trying to use sfcli.py to do a search. I should have been more precise with my PR comment. In any case, I can easily define a valid input type for this function that is not a dict (prior to this change, of course).

bcoles: Looks like we were both looking at `sfdb` about the same time. I PR'd some unit tests the same day as #604, but had the `search` tests commented out until I could fix it. You beat me to it.

bcoles: This should be mostly fixed in #610 which ensures that sfdb raises errors (in most instances). There's still a bit more cleaning up to be done.

PR #608, opened 2020-06-26
sfscan: move startScan() functionality out of init()

### LINKED ISSUE #554: Error with module sfp_abusech.py
Hi, 
I just installed the spiderfoot. I install it like this :  
`git clone  https://github.com/smicallef/spiderfoot.git ` after that I did : `pip3 install -r requierement.txt
`
And for try you tool I follow you video and I did this : 
`python3 sf.py -m sfr_abusech.py -s <IP> -q` 

I have this error : 
```
SpiderFoot UI                   IP Address                                      <IP>
[Fatal] Unable to create instance in DB: UNIQUE constraint failed: tbl_scan_instance.guid
[FrameInfo(frame=<frame at 0x1fb0e88, file '/root/Documents/Test/spiderfoot/sflib.py', line 328, code fatal>, filename='/root/Documents/Test/spiderfoot/sflib.py', lineno=328, function='fatal', code_context=['        print(str(inspect.stack()))\n'], index=0), FrameInfo(frame=<frame at 0x1a084b8, file '/root/Documents/Test/spiderfoot/sfdb.py', line 408, code scanInstanceCreate>, filename='/root/Documents/Test/spiderfoot/sfdb.py', lineno=408, function='scanInstanceCreate', code_context=['                self.sf.fatal("Unable to create instance in DB: " + e.args[0])\n'], index=0), FrameInfo(frame=<frame at 0x1f84d88, file '/root/Documents/Test/spiderfoot/sfscan.py', line 80, code startScan>, filename='/root/Documents/Test/spiderfoot/sfscan.py', lineno=80, function='startScan', code_context=['                                       self.scanName, self.targetValue)\n'], index=0), FrameInfo(frame=<frame at 0x7fdc89559388, file '/root/Documents/Test/spiderfoot/sfscan.py', line 54, code run>, filename='/root/Documents/Test/spiderfoot/sfscan.py', lineno=54, function='run', code_context=['        self.startScan()\n'], index=0), FrameInfo(frame=<frame at 0xb61b78, file 'sf.py', line 331, code <module>>, filename='sf.py', lineno=331, function='<module>', code_context=['        t.run()\n'], index=0)]
```
I forgot something in the installation ? I have only ssh access I don't use GUI.

bcoles:
This should also fix #554 by ensuring scan IDs are automatically generated.

### DIFF
--- a/sf.py
+++ b/sf.py
@@ -320,7 +320,6 @@ if __name__ == '__main__':
         if sfConfig['__logging']:
             print(("[*] Modules enabled (" + str(len(modlist)) + "): " + ",".join(modlist)))
         cfg = sf.configUnserialize(dbh.configGet(), sfConfig)
-        scanId = sf.genScanInstanceGUID(target)
 
         # Debug mode is a variable that gets stored to the DB, so re-apply it
         if args.debug:
@@ -332,10 +331,20 @@ if __name__ == '__main__':
         if args.x and args.t:
             cfg['__outputfilter'] = args.t.split(",")
 
-        t = SpiderFootScanner(target, target, targetType, scanId,
-            modlist, cfg, dict())
-        t.daemon = True
-        t.run()
+        try:
+            s = SpiderFootScanner(target, target, targetType, scanId, modlist, cfg)
+            scanId = s.getId()
+        except BaseException as e:
+            print("[-] Failed to initialize scan: %s" % e)
+            sys.exit(-1)
+
+        try:
+            p = mp.Process(target=s.startScan())
+            p.daemon = True
+            p.start()
+        except BaseException as e:
+            print("[-] Scan [%s] failed: %s" % (scanId, e))
+            sys.exit(-1)
 
         # If field headers weren't disabled, print them
         if not args.H and args.o != "json":
--- a/sflib.py
+++ b/sflib.py
@@ -369,11 +369,8 @@ class SpiderFoot:
         """Set the GUID this instance of SpiderFoot is being used in."""
         self.GUID = uid
 
-    def genScanInstanceGUID(self, scanName):
+    def genScanInstanceGUID(self):
         """Generate an globally unique ID for this scan.
-
-        Args:
-            scanName (str): scan name
 
         Returns:
             str: scan instance unique GUID
@@ -636,10 +633,24 @@ class SpiderFoot:
     # Configuration process
     #
 
-    # Convert a Python dictionary to something storable
-    # in the database.
     def configSerialize(self, opts, filterSystem=True):
+        """Convert a Python dictionary to something storable in the database.
+
+        Args:
+            opts (dict): TBD
+            filterSystem (bool): TBD
+
+        Returns:
+            dict: config options
+        """
+
+        if not isinstance(opts, dict):
+            raise TypeError("opts is %s; expected dict()" % type(opts))
+
         storeopts = dict()
+
+        if not opts:
+            return storeopts
 
         for opt in list(opts.keys()):
             # Filter out system temporary variables like GUID and others
@@ -657,7 +668,7 @@ class SpiderFoot:
             if type(opts[opt]) is list:
                 storeopts[opt] = ','.join(opts[opt])
 
-        if '__modules__' not in opts:
+        if opts['__modules__'] is None:
             return storeopts
 
         for mod in opts['__modules__']:
@@ -680,11 +691,26 @@ class SpiderFoot:
 
         return storeopts
 
-    # Take strings, etc. from the database or UI and convert them
-    # to a dictionary for Python to process.
-    # referencePoint is needed to know the actual types the options
-    # are supposed to be.
     def configUnserialize(self, opts, referencePoint, filterSystem=True):
+        """Take strings, etc. from the database or UI and convert them
+        to a dictionary for Python to process.
+        referencePoint is needed to know the actual types the options
+        are supposed to be.
+
+        Args:
+            opts (dict): TBD
+            referencePoint (dict): TBD
+            filterSystem (bool): TBD
+
+        Returns:
+            dict: TBD
+        """
+
+        if not isinstance(opts, dict):
+            raise TypeError("opts is %s; expected dict()" % type(opts))
+        if not isinstance(referencePoint, dict):
+            raise TypeError("referencePoint is %s; expected dict()" % type(referencePoint))
+
         returnOpts = referencePoint
 
         # Global options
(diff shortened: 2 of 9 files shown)


sflib.py
```diff
@@ -657,7 +668,7 @@ def configSerialize(self, opts, filterSystem=True):
             if type(opts[opt]) is list:
                 storeopts[opt] = ','.join(opts[opt])
 
-        if '__modules__' not in opts:
+        if opts['__modules__'] is None:
```
smicallef, line 671: This will generate an exception if `__modules__` doesn't exist. I haven't looked thoroughly but there may be cases where it doesn't exist, so would be better to leave as it was or use `.get()` instead.

sf.py
```diff
@@ -332,10 +331,20 @@ def handle_abort(signal, frame):
         if args.x and args.t:
             cfg['__outputfilter'] = args.t.split(",")
 
-        t = SpiderFootScanner(target, target, targetType, scanId,
-            modlist, cfg, dict())
-        t.daemon = True
-        t.run()
+        try:
+            s = SpiderFootScanner(target, target, targetType, scanId, modlist, cfg)
+            scanId = s.getId()
```
smicallef, line 336: Possible race: at this stage the ID might not be available.

bcoles: Disagree. The ID is set during `SpiderFootScanner` initialisation. This line is called after the `SpiderFootScanner` object is created.

sfwebui.py
```diff
@@ -363,27 +362,35 @@ def rerunscan(self, id):
             scantarget = scantarget.lower()
 
         # Start running a new scan
-        newId = sf.genScanInstanceGUID(scanname)
-        p = mp.Process(target=SpiderFootScanner, args=(scanname, scantarget, targetType, newId,
-                modlist, cfg, modopts))
-        p.start()
+        try:
+            s = SpiderFootScanner(scanname, scantarget, targetType, modlist, cfg)
+            scanId = s.getId()
```
smicallef, line 367: Race condition mentioned above.

bcoles: Disagree. The ID is set during SpiderFootScanner initialisation. This line is called after the SpiderFootScanner object is created.

sfwebui.py
```diff
@@ -409,13 +416,23 @@ def rerunscanmulti(self, ids):
                                   "a human name, IP address, IP subnet, ASN, domain name or host name.")
 
             # Start running a new scan
-            newId = sf.genScanInstanceGUID(scanname)
-            p = mp.Process(target=SpiderFootScanner, args=(scanname, scantarget.lower(), 
-                            targetType, newId, modlist, cfg, modopts))
-            p.start()
+            try:
+                s = SpiderFootScanner(scanname, scantarget, targetType, modlist, cfg)
+                s.getId()
```
smicallef, line 421: Race

bcoles: Disagree. The ID is set during SpiderFootScanner initialisation. This line is called after the SpiderFootScanner object is created.

sfwebui.py
```diff
@@ -835,15 +851,25 @@ def startscan(self, scanname, scantarget, modulelist, typelist, usecase, cli=Non
...
         else:
             scantarget = scantarget.lower()
-        p = mp.Process(target=SpiderFootScanner, args=(scanname, scantarget, targetType, scanId,
-                              modlist, cfg, modopts))
-        p.daemon = True
-        p.start()
+
+        try:
+            s = SpiderFootScanner(scanname, scantarget, targetType, modlist, cfg)
+            scanId = s.getId()
```
smicallef, line 861: Race

bcoles: Disagree. The ID is set during SpiderFootScanner initialisation. This line is called after the SpiderFootScanner object is created.

sfwebui.py
bcoles: However, there is an error with this line. It's supposed to be `scanId = s.getId()`

sflib.py
bcoles: Good. It should error here. Better error handling would be great, but I would rather it die here due to a missing key rather than die several layers deeper resulting in an unintuitive stack trace like this (which is what happens with the original code or when using `get()`):

```
_____________________________________________________________ TestSpiderFootScanner.test_set_status_invalid_status_should_raise ______________________________________________________________

self = <test.unit.test_spiderfootscanner.TestSpiderFootScanner testMethod=test_set_status_invalid_status_should_raise>

    def test_set_status_invalid_status_should_raise(self):
        """
        Test def setStatus(self, status, started=None, ended=None)
        """
>       sfscan = SpiderFootScanner("", "", "IP_ADDRESS", list(), self.default_options)

test/unit/test_spiderfootscanner.py:86: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sfscan.py:82: in __init__
    self.dbh.scanConfigSet(self.scanId, self.sf.configSerialize(deepcopy(self.config)))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sflib.SpiderFoot object at 0x7f608b0df280>, opts = {'__blocknotif': False, '__database': 'spiderfoot.test.db', '__docroot': '', '__logging': True, ...}, filterSystem = True

    def configSerialize(self, opts, filterSystem=True):
        """Convert a Python dictionary to something storable in the database.
    
        Args:
            opts (dict): TBD
            filterSystem (bool): TBD
    
        Returns:
            dict: config options
        """
    
        if not isinstance(opts, dict):
            raise TypeError("opts is %s; expected dict()" % type(opts))
    
        storeopts = dict()
    
        if not opts:
            return storeopts
    
        for opt in list(opts.keys()):
            # Filter out system temporary variables like GUID and others
            if opt.startswith('__') and filterSystem:
                continue
    
            if type(opts[opt]) is int or type(opts[opt]) is str:
                storeopts[opt] = opts[opt]
    
            if type(opts[opt]) is bool:
                if opts[opt]:
                    storeopts[opt] = 1
                else:
                    storeopts[opt] = 0
            if type(opts[opt]) is list:
                storeopts[opt] = ','.join(opts[opt])
    
>       if not opts.get(['__modules__']):
E       TypeError: unhashable type: 'list'

sflib.py:671: TypeError
```

bcoles:
```suggestion
        # todo: ensure the __modules__ key value is in the expected format
        if '__modules__' not in opts:
```

PR #611, opened 2020-06-27
tests: update sfdb and sfwebui tests

bcoles:


sfdb.py
```diff
@@ -574,7 +585,9 @@ def scanInstanceGet(self, instanceId):
         with self.dbhLock:
             try:
                 self.dbh.execute(qry, qvars)
-                return self.dbh.fetchone()
+                ret = self.dbh.fetchone()
+                if not ret:
+                    return list()
```
bcoles, line 590: This was horribly misguided and should be reverted.

```suggestion
                return self.dbh.fetchone()
```

PR #627, opened 2020-07-04
Add sfp_jsonwhoisio

bcoles:
I have no idea if this works.

The `availability` API endpoint is access to trial accounts, but appears to always return `null` rather than valid JSON. Perhaps it is broken. The web interface shows that the queries were processed successfully.

The domain WHOIS API endpoint requires a paid account which I don't have, so I tested with a mocked up response on a test server using test data from the API documentation:

* https://jsonwhois.io/docs/


smicallef:
Still waiting on a response from someone there to help with the integration testing :(

smicallef:
Still no word from them, and based on their website even the demo is returning an error. Perhaps time to abandon this?

PR #630, opened 2020-07-06
Added links to API keys management

dafyk:
API registration/management links for multiple modules description. Not all modules can be linked (e.g. VirusTotal uses username in link)


smicallef:
Thanks for the PR! Unfortunately having HTML in the option descriptions is not really the direction I'd like to go, primarily because the option descriptions may also be visible outside of the web UI where HTML would make the option difficult to read.

We'll soon be putting out a new module structure that will include a configuration definition that's much cleaner than the current setup, so there will then be the opportunity to include these links and have them cleanly handled by the UI.

I'll go ahead and close this, but your contribution was definitely appreciated!

PR #637, opened 2020-07-10
sf: Fix JSON output formatting when running sf from command line

### LINKED ISSUE #636: error in json structure
The json structure produced in output, from the latest versions of Spiderfoot, does not seem correct.

**This is the command used:**
`sf.py -m sfp_abusech,sfp_abuseipdb,sfp_alienvault,sfp_alienvaultiprep,sfp_greensnowsfp_badipscom,sfp_blocklistde,sfp_botscout.py,sfp_cinsscore,sfp_cleantalk,sfp_cybercrimetracker,sfp_isc,sfp_nothink,sfp_fortinet,sfp_talosintel,sfp_threatcrowd,sfp_torexits,sfp_totalhash,sfp_virustotal,sfp_sorbs -s 217.28.98.112 -o json`

**This is the output produced with version 3.0:**
[{"generated": 1594320400, "type": "IP Address", "data": "217.28.98.112", "module": "SpiderFoot UI", "source": "217.28.98.112"}]

**This is the output produced with the latest version available today:**
{"generated": 1594319181, "type": "IP Address", "data": "217.28.98.112", "module": "SpiderFoot UI", "source": "217.28.98.112"}[]

bcoles:
Obviously having a new line after `[` is not ideal, but I'm not sure how else to fix this? Especially if both Windows and *nix need to be supported.

Fix #636

### DIFF
--- a/sf.py
+++ b/sf.py
@@ -373,7 +373,7 @@ if __name__ == '__main__':
                     print('{0:30}{1}{2:45}{3}{4}{5}{6}'.format("Source", delim, "Type", delim, "Source Data", delim, "Data"))
 
         if args.o == "json":
-            print("[", end='')
+            print("[")
 
         while True:
             info = dbh.scanInstanceGet(scanId)


steve-bate:
I see the same behavior on Linux and Windows. It seems to be happening because the scan is run asynchronously (in a subprocess) before the opening bracket is printed. The makes the bracket ordering nondeterministic, but it will usually be after the first log entry on a fast machine. I haven't researched the issue enough to determine the best fix, but one workaround is to print the opening bracket *before* launching the scan subprocess.

bcoles:
I've created another PR #639 but this also isn't ideal.

PR #639, opened 2020-07-10
sf: Fix JSON output formatting when running sf from command line

### LINKED ISSUE #636: error in json structure

bcoles:
This isn't ideal, as the `[` is printed even when the scan fails due to an error.

Fix #636

### DIFF
--- a/sf.py
+++ b/sf.py
@@ -339,6 +339,9 @@ if __name__ == '__main__':
         if args.x and args.t:
             cfg['__outputfilter'] = args.t.split(",")
 
+        if args.o == "json":
+            print("[", end='')
+
         # Start running a new scan
         scanName = target
         scanId = sf.genScanInstanceGUID()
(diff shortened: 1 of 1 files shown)


steve-bate:
Agreed. Just brainstorming... in command line mode, is it possible to not print the log entries as they are being produced and, instead, query the database and print the resulting JSON at the end of the run? If the goal is to have the log entries stream to stdout during the scan, it's tricky with multiple processes writing to stdout and using JSON.

bcoles:
> query the database and print the resulting JSON at the end of the run?

Polling periodically may be another alternative.

> If the goal is to have the log entries stream to stdout during the scan, it's tricky with multiple processes writing to stdout and using JSON.

That would be preferred approach. I'm not interested in reviewing this further for now.

smicallef:
Merging for now but will re-visit to find a better way to handle this.

PR #693, opened 2020-08-10
sfp_emailrep: Add support for API key

### LINKED ISSUE #688: EmailRep.io doesn't support an API key
Looking at the config settings for emailrep.io there currently aren't any. Would it be possible to add support for an API key?

bcoles:
I have no idea if this works. I haven't tested with a key.

Fixes #688


smicallef:
I've requested a key, so will sit on this a few days until I get the key and am able to test it out.

PR #724, opened 2020-08-16
Added meta for abuse.ch

krishnasism:


modules/sfp_abusech.py
```diff
@@ -58,6 +58,58 @@
...
+
+            'model': "FREE_AUTH_UNLIMITED",
+
+            'references': [
+                "https://bazaar.abuse.ch/",
+                "https://feodotracker.abuse.ch/",
+                "https://igotphished.abuse.ch/",
+                "https://sslbl.abuse.ch/",
+                "https://urlhaus.abuse.ch/",
+                "https://www.datasource.com/api-documentation"
```
smicallef, line 83: That can be removed, it's just an example.

modules/sfp_abusech.py
```diff
@@ -58,6 +58,58 @@
 }
 
 class sfp_abusech(SpiderFootPlugin):
+meta = {
+        'name': "abuse.ch",
+
```
smicallef, line 63: No spaces between each line

modules/sfp_abusech.py
```diff
@@ -58,8 +58,45 @@
 }
 
 class sfp_abusech(SpiderFootPlugin):
-    """abuse.ch:Investigate,Passive:Reputation Systems::Check if a host/domain, IP or netblock is malicious according to abuse.ch."""
```
smicallef, line 61: This should remain in place for now. We will only remove it once the meta objects are fully in place everywhere and we've updated the code that relies upon its existence.

PR #732, opened 2020-08-17
Update arin meta

krishnasism:


modules/sfp_arin.py
```diff
@@ -23,7 +23,28 @@ class sfp_arin(SpiderFootPlugin):
         'summary': "Queries ARIN registry for contact information.",
         'flags': [ "" ],
         'useCases': [ "Footprint", "Investigate", "Passive" ],
-        'categories': [ "Public Registries" ]
+        'categories': [ "Public Registries" ],
+        'dataSource': {
+            'website': "https://www.arin.net/",
+            'model': "COMMERCIAL_ONLY",
```
smicallef, line 29: No API key needed so not commercial.

PR #813, opened 2020-08-19
Update opencorporates meta

krishnasism:


modules/sfp_opencorporates.py
```diff
@@ -18,12 +18,25 @@ class sfp_opencorporates(SpiderFootPlugin):
...
-		'categories': [ "Search Engines" ]
-	}
+        'name': "OpenCorporates",
+        'summary': "Look up company information from OpenCorporates.",
+        'flags': [ "" ],
+        'useCases': [ "Passive", "Footprint", "Investigate" ],
+        'categories': [ "Search Engines" ],
+        'dataSource': {
+            'website': "https://opencorporates.com",
+            'model': "FREE_NOAUTH_UNLIMITED",
```
smicallef, line 28: I think it's FREE_NOAUTH_LIMITED since you need to upgrade to get higher query volumes.

krishnasism: Updated

PR #817, opened 2020-08-19
Update pastebin meta

krishnasism:


modules/sfp_pastebin.py
```diff
@@ -20,12 +20,32 @@ class sfp_pastebin(SpiderFootPlugin):
...
+            'model': "FREE_AUTH_LIMITED",
+            'references': [
+                "https://pastebin.com/doc_api",
+                "https://pastebin.com/faq"
+            ],
+            'apiKeyInstructions': [
+                "Visit pastebin.com",
+                "Register a free account",
+                "Navigate to pastebin.com/doc_api",
+                "The API key is listed under 'Your Unique Developer API Key'"
```
smicallef, line 39: Actually it's the Google search API they need to obtain.

krishnasism: So, instructions for getting google custom search api?

smicallef: Yep

krishnasism: Updated

PR #837, opened 2020-08-20
Update spur meta

krishnasism:


modules/sfp_spur.py
```diff
@@ -19,12 +19,25 @@ class sfp_spur(SpiderFootPlugin):
...
+        'summary': "Obtain information about any malicious activities involving IP addresses found",
+        'flags': [ "apikey" ],
+        'useCases': [ "Investigate", "Passive" ],
+        'categories': [ "Reputation Systems" ],
+        'dataSource': {
+            'website': "https://spur.us/",
+            'model': "COMMERCIAL ONLY",
+            'references': [
+                "https://spur.us/api"
+            ],
```
smicallef, line 32: Missing instructions?

krishnasism: Couldn't go past the paywall, would require your input

smicallef: ok -

1. Pay for a subscription
2. Under the Context API menu on the left, click Tokens
3. Your API key is presented to you

krishnasism: Thanks, updated

PR #849, opened 2020-08-20
Update twilio meta

krishnasism:


modules/sfp_twilio.py
```diff
@@ -17,13 +17,31 @@
 class sfp_twilio(SpiderFootPlugin):
     """Twilio:Footprint,Investigate,Passive:Search Engines:apikey:Obtain information from Twilio about phone numbers. Ensure you have the Caller Name add-on installed in Twilio."""
     
-    meta = {
```
smicallef, line 20: Extra line deleted :)

krishnasism: Yeah, just pushed again

PR #850, opened 2020-08-20
Update twitter meta

krishnasism:


modules/sfp_twitter.py
```diff
@@ -17,12 +17,26 @@ class sfp_twitter(SpiderFootPlugin):
...
-	}
+        'name': "Twitter",
+        'summary': "Gather name and location from Twitter profiles.",
+        'flags': [ "" ],
+        'useCases': [ "Footprint", "Investigate", "Passive" ],
+        'categories': [ "Social Media" ],
+        'dataSource': {
+            'website': "https://twitter.com/",
+            'model': "FREE_NOAUTH_UNLIMITED",
+            'references': [
```
smicallef, line 28: As this module doesn't use the API, no need to link to API reference docs.

PR #913, opened 2020-08-24
Add Host.io query module

leotrubach:


modules/sfp_hostio.py
```diff
@@ -0,0 +1,179 @@
...
+            evt = SpiderFootEvent('IP_ADDRESS', address, self.__name__, event)
+            self.notifyListeners(evt)
+
+            country = ip_data.get("country")
+            if country is not None:
+                country_evt = SpiderFootEvent("COUNTRY_NAME", country, self.__name__, evt)
+                self.notifyListeners(country_evt)
+
+        evt = SpiderFootEvent("RAW_RIR_DATA", json.dumps(data), self.__name__, event)
+        self.notifyListeners(evt)
```
smicallef, line 179: Newline missing at end of file.

modules/sfp_hostio.py
```diff
@@ -0,0 +1,179 @@
...
+            "AFFILIATE_DOMAIN_NAME",
+            "CO_HOSTED_SITE_DOMAIN",
+            "DOMAIN_NAME",
+        ]
+
+    # What events this module produces
+    def producedEvents(self):
+        return [
+            "IP_ADDRESS",
+            "COUNTRY_NAME"
```
smicallef, line 81: RAW_RIR_DATA is not declared here.

modules/sfp_hostio.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def watchedEvents(self):
+        return [
+            "AFFILIATE_DOMAIN_NAME",
+            "CO_HOSTED_SITE_DOMAIN",
+            "DOMAIN_NAME",
+        ]
+
+    # What events this module produces
+    def producedEvents(self):
```
smicallef, line 78: The module could get more from their API, like EMAILADDR, WEB_ANALYTICS_ID and WEBSERVER_TECHNOLOGY.

modules/sfp_hostio.py
```diff
@@ -0,0 +1,179 @@
...
+        if userOpts is None:
+            userOpts = {}
+        self.sf = sfc
+        self.results = self.tempStorage()
+        self.opts.update(userOpts)
+
+    def watchedEvents(self):
+        return [
+            "AFFILIATE_DOMAIN_NAME",
+            "CO_HOSTED_SITE_DOMAIN",
```
smicallef, line 73: There's little value in watching for AFFILIATE_DOMAIN_NAME and CO_HOSTED_SITE_DOMAIN for this module, so I suggest you drop both.

modules/sfp_hostio.py
```diff
@@ -0,0 +1,179 @@
...
+        "verify": True,
+    }
+
+    optdescs = {
+        "api_key": "Host.io API Key.",
+        "checkcohosts": "Check co-hosted sites?",
+        "checkaffiliates": "Check affiliates?",
+        "maxcohost": "Stop reporting co-hosted sites after this many are found, as it would likely indicate web hosting.",
+        "verify": "Verify that any hostnames found on the target domain still resolve?",
+    }
```
smicallef, line 59: The above three config options aren't used, so they can be removed from both optdescs and opts.

PR #922, opened 2020-08-25
Remove duplicate response data

faleksic:


modules/sfp_onyphe.py
```diff
@@ -260,6 +272,9 @@ def handleEvent(self, event):
                     )
                     self.sf.info("Found GeoIP for " + eventData + ": " + location)
 
+                    if self.isDuplicate(location):
```
smicallef, line 275: This approach is not really effective since the same location may appear multiple times but for different data sources, and that's ok. What is not ok is when the same data is reported multiple times for the same data source.

PR #924, opened 2020-08-26
Add title, coordinates and geoinfo to sfp_hostio

leotrubach:


modules/sfp_hostio.py
```diff
@@ -149,7 +152,7 @@ def handleEvent(self, event):
             for address, ip_data in data["ipinfo"].items():
                 evt = SpiderFootEvent("IP_ADDRESS", address, self.__name__, event)
                 self.notifyListeners(evt)
-                fonud = True
+                found = True
```
smicallef, line 155: Good spot!

PR #951, opened 2020-08-30
flake8: resolve flake8 violations: E121

bcoles:
* Resolve flake8 violations: E121
* Adds docstrings to functions in Spyse module.
* Changes a few modules to use f-strings.
* Adds a `# TODO: review, is this key meant to be 'operating_system' ?` comment to the spyse module


smicallef:
In answer to the TODO, no that is actually the name of the field in their API.

PR #959, opened 2020-08-31
Implement BitcoinAbuse functionality

leotrubach:


modules/sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,134 @@
...
+        srcModuleName = event.module
+        eventData = event.data
+
+        if self.errorState:
+            return None
+
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        if self.opts["api_key"] == "":
+            self.sf.error("You enabled sfp_template but did not set an API key!", False)
```
bcoles, line 110:
```suggestion
            self.sf.error("You enabled sfp_bitcoinabuse but did not set an API key!", False)
```

modules/sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,134 @@
...
+            self.errorState = True
+            return None
+
+        if eventData in self.results:
+            self.sf.debug(f"Skipping {eventData}, already checked.")
+            return None
+        else:
+            self.results[eventData] = True
```
bcoles, line 118: Redundant `else` after `return`.

```suggestion

        self.results[eventData] = True
```

test/unit/modules/test_sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,86 @@
...
+        '_socks2addr': '',
+        '_socks3port': '',
+        '_socks4user': '',
+        '_socks5pwd': '',
+        '_socks6dns': True,
+        '_torctlport': 9051,
+        '__logstdout': False
+    }
+
+    @unittest.skip("todo")
```
bcoles, line 41: Should be safe to remove:

```suggestion
```

modules/sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if eventName == "BITCOIN_ADDRESS":
+            rec = self.query(eventData)
+            if isinstance(rec, dict):
+                count = rec.get("count")
+                if isinstance(count, int):
+                    if count != 0:
```
bcoles, line 125: A `count` value of zero should evaluate to false.

```suggestion
                    if count:
```

or, of negative values are possible:

```suggestion
                    if count > 0:
```

modules/sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,134 @@
...
+        self.sf = sfc
+        self.results = self.tempStorage()
+
+        if userOpts:
+            self.opts.update(userOpts)
+
+    def watchedEvents(self):
+        return ["BITCOIN_ADDRESS"]
+
+    # What events this module produces
```
bcoles, line 71:
```suggestion
```

modules/sfp_bitcoinabuse.py
leotrubach: Made same changes to sfp_template

modules/sfp_bitcoinabuse.py
```diff
@@ -0,0 +1,133 @@
...
+        self.results[eventData] = True
+
+        if eventName == "BITCOIN_ADDRESS":
+            rec = self.query(eventData)
+            if isinstance(rec, dict):
+                count = rec.get("count")
+                if isinstance(count, int):
+                    if count > 0:
+                        evt = SpiderFootEvent(
+                            "MALICIOUS_BITCOIN_ADDRESS", rec["address"], self.__name__, event
```
smicallef, line 126: Is there any additional text that can go in the result to make it more meaningful? At the very least it should be modeled around the other modules that report malicious entities (see `sfp_emailrep.py` as an example).

leotrubach: For bitcoinabuse there is `count` parameter that can be included. 
@smicallef what do you think about serializing that as JSON data (like JSON object of defined schema)? In that case it will become the same as RIR data though. 
There might be some additional fields after we implement Whoswho functionality.

leotrubach:
```
```

bcoles: > For bitcoinabuse there is `count` parameter that can be included.
> @smicallef what do you think about serializing that as JSON data (like JSON object of defined schema)? In that case it will become the same as RIR data though.
> There might be some additional fields after we implement Whoswho functionality.

The raw data is already logged in `RAW_RIR_DATA`.

> ```
> credentials_leaked = details.get('credentials_leaked')
>         if credentials_leaked:
>             evt = SpiderFootEvent('EMAILADDR_COMPROMISED', eventData + " [Unknown]", self.__name__, event)
>             self.notifyListeners(evt)
> 
>         malicious_activity = details.get('malicious_activity')
>         if malicious_activity:
>             evt = SpiderFootEvent('MALICIOUS_EMAILADDR', 'EmailRep [' + eventData + ']', self.__name__, event)
>             self.notifyListeners(evt)
> ```
> 
> I see the following code in sfp_emailrep.

Something like that (`eventData + " [BitcoinAbuse]"`) will do for now. I presume smicallef meant something like that.

> Do you think it is good enough? I mean, other modules need to parse it, write REGEXes. With JSON data it could be cleaner approach.

Yeah, I know, it's a dumb way to format data. This format sucks, but at least it sucks consistently.

I'm not convinced that using JSON is a better solution either. That doesn't fit with the existing architecture.

My opinion. Obviously it's @smicallef's decision.

smicallef: Agree it's dumb but like @bcoles said, consistently dumb is better, and we don't have a good alternative at the moment.

PR #963, opened 2020-09-01
Add module sfp_c99

faleksic:


modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def producedEvents(self):
+        return [
+            "RAW_RIR_DATA",
+            "GEOINFO",
+            "INTERNET_NAME",
+            "INTERNET_NAME_UNRESOLVED",
+            "PROVIDER_TELCO",
+            "PHYSICAL_ADDRESS",
+            "PHYSICAL_COORDINATES" "PROVIDER_DNS",
```
smicallef, line 81: Missing comma?

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+# Copyright:   (c) Steve Micallef
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import json
+from sflib import SpiderFootPlugin, SpiderFootEvent
+
+
+class sfp_c99(SpiderFootPlugin):
+    """C99:Footprint,Passive,Investigate:Content Analysis:apikey:Obtain data from c99 API which offers various data (geo location, proxy detection, phone lookup, etc)."""
```
smicallef, line 18: Docstring is no longer needed.

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+        "categories": ["Content Analysis"],
+        "dataSource": {
+            "website": "https://api.c99.nl/",
+            "model": "COMMERCIAL_ONLY",
+            "references": ["https://api.c99.nl/api_overview", "https://api.c99.nl/faq"],
+            "apiKeyInstructions": [
+                "Visit api.c99.nl",
+                "Press shop in navigation or go to https://api.c99.nl/dashboard/shop",
+                "Press purchase key on option 'C99.NL API KEY' (you can also buy 1 year key if you want)",
+                "You should get your key on your email'",
```
smicallef, line 34: "You will receive your API key by email." Also note the hanging single quote at the end.

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+            "logo": "https://api.c99.nl/assets/images/logo.png",
+            "description": "C99 API service is versatile source of information. "
+            "They offer over 57 different APIs of which 10 are integrated in this module. "
+            "APIs that are integrated are subdomain finder, phone lookup, Skype resolver, "
+            "IP to Skype, firewall technology WAF detector, domain history, "
+            "IP to domains, IP geo location, proxy detector.",
+        },
+    }
+
+    opts = {"api_key": "", "response_type": "JSON", "verify": True}
```
smicallef, line 46: Any reason why we care about the response_type? Why not always have it be JSON?

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+            return None
+
+        if (
+            self.opts["response_type"].lower() != "plain"
+            and self.opts["response_type"].lower() != "json"
+        ):
+            self.sf.info(
+                "You enabled sfp_c99 ,but did not set response type to json or plain! JSON will be used by default"
+            )
+            self.opts["response_type"] = "json"
```
smicallef, line 556: As per my earlier comment, I suggest we stick with JSON and drop all the logic associated with `response_type`.

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+            self.sf.debug(f"Skipping {eventData}, already checked.")
+            return None
+
+        self.results[eventData] = True
+
+        if eventName == "PHONE_NUMBER":
+            phoneData = self.query("phonelookup", "number", eventData)
+            phoneData = phoneData.get("details") if phoneData is not None else None
+
+            if phoneData is not None:
```
smicallef, line 569: I think it's safer to use `if phoneData` vs. `if phoneData is not None` for two reasons: 1) Empty strings, booleans and empty lists will not be `None` and evaluate as `True` here, so that might have unexpected consequences. 2) IMHO it's easier to read.

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+            if isinstance(domainHistoryData, list):
+                self.emitDomainHistoryData(domainHistoryData, event)
+
+            wafDetectorData = self.query("firewalldetector", "url", eventData)
+
+            if wafDetectorData is not None:
+                self.emitWafDetectorData(wafDetectorData, event)
+
+        if eventName == "IP_ADDRESS":
+            IPToSkypeData = self.query("ip2skype", "ip", eventData)
```
smicallef, line 597: Whilst I understand the use of IP in the name here, normally class names start with upper case, so better to use `ip` instead of `IP`.

modules/sfp_c99.py
```diff
@@ -0,0 +1,624 @@
...
+        ):
+            self.sf.debug(f"Host {hostName} could not be resolved for {event.data}")
+            evt = SpiderFootEvent(
+                "INTERNET_NAME_UNRESOLVED",
+                hostName,
+                self.__name__,
+                event,
+            )
+            self.notifyListeners(evt)
+        else:
```
smicallef, line 430: Based on the `if` clause, this will be executed when `hostName` is None, which I don't think is intended?

modules/sfp_c99.py
faleksic: Definitely agree 👍

modules/sfp_c99.py
faleksic: They offer two types of responses so I thought it might be good idea to give user an option to choose, but after implementing it I can see it doesn't make sense. JSON response gives so much more data and it is simpler to use

PR #992, opened 2020-09-06
Implement ZoneCruncher functionality

leotrubach:


modules/sfp_zonecruncher.py
```diff
@@ -0,0 +1,186 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         sfp_zonecruncher
```
smicallef, line 3: Why zone cruncher instead of Zetalytics?

test/unit/modules/test_sfp_zonecruncher.py
```diff
@@ -0,0 +1,85 @@
+# test_sfp_zonecruncher.py
+import unittest
+
+from modules.sfp_zonecruncher import sfp_zonecruncher
+from sflib import SpiderFoot
+from spiderfoot import SpiderFootEvent, SpiderFootTarget
+
+
+class TestModuleBitcoinAbuse(unittest.TestCase):
```
smicallef, line 9: Wrong test naming. Also reflected below.

modules/sfp_zonecruncher.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if self.opts["api_key"] == "":
+            self.sf.error("You enabled sfp_template but did not set an API key!", False)
+            self.errorState = True
+            return None
+
+        if (eventName, eventData) in self.results:
+            self.sf.debug(f"Skipping {eventName}:{eventData}, already checked.")
+            return None
+        self.results[eventName, eventData] = True
```
smicallef, line 166: Please make the key a string instead of a tuple, as there is likely code elsewhere (or HX) that relies on this being a string for now.

modules/sfp_zetalytics.py
```diff
@@ -0,0 +1,186 @@
...
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_zetalytics(SpiderFootPlugin):
+    BASE_URL = "https://zonecruncher.com/api/v1"
+    meta = {
+        "name": "ZoneCruncher",
+        "summary": "Query ZoneCruncher database",
+        "flags": ["apikey"],
```
smicallef, line 24: The above plus various other points still make reference to Zonecruncher.

modules/sfp_zetalytics.py
```diff
@@ -0,0 +1,186 @@
...
+        "useCases": ["Passive"],
+        "categories": ["Passive DNS"],
+        "dataSource": {
+            "website": "https://zetalytics.com/",
+            "model": "FREE_AUTH_LIMITED",
+            "references": ["https://zonecruncher.com/api-v1-docs/"],
+            "apiKeyInstructions": [
+                "Visit https://mailchi.mp/zetalytics/trial-access-request",
+                "Register a free account and request API key",
```
smicallef, line 33: request an API key

test/unit/modules/test_sfp_zetalytics.py
```diff
@@ -0,0 +1,85 @@
...
+import unittest
+
+from modules.sfp_zetalytics import sfp_zetalytics
+from sflib import SpiderFoot
+from spiderfoot import SpiderFootEvent, SpiderFootTarget
+
+
+class TestModuleZetalytics(unittest.TestCase):
+    """
+    Test modules.sfp_bitcoinabuse
```
smicallef, line 11: Still mentions bitcoinabuse

bcoles:
I haven't looked at the format of the response data, but I think it would make more sense to perform `query_subdomains` for `DOMAIN_NAME` rather than `INTERNET_NAME` (?)

All `INTERNET_NAME` events are target children of a `DOMAIN_NAME` event. Presumably a query for subdomains on the `DOMAIN_NAME` should identify all `INTERNET_NAME` subdomains in one query, rather than querying for every `INTERNET_NAME` individually, unless there's some sort of response length restrictions.

```python
        if eventName == "INTERNET_NAME":
            data = self.query_subdomains(eventData)
            self.generate_subdomains_events(data, event)

            data = self.query_hostname(eventData)
            self.generate_hostname_events(data, event)

        elif eventName == "DOMAIN_NAME":
            data = self.query_email_domain(eventData)
            self.generate_email_domain_events(data, event)
```

bcoles:
This module is missing some code patterns for determining whether the identified data is within the target scope using `self.getTarget()`. Here's an example:

```python
        for domain in set(domains):
            if self.checkForStop():
                return None
                
            if domain in self.results:
                continue

            if not self.getTarget().matches(domain, includeChildren=True):
                continue

            if self.opts['verify'] and not self.sf.resolveHost(domain):
                self.sf.debug(f"Host {domain} could not be resolved")
                evt = SpiderFootEvent("INTERNET_NAME_UNRESOLVED", domain, self.__name__, event)
                self.notifyListeners(evt)
            else:
                evt = SpiderFootEvent("INTERNET_NAME", domain, self.__name__, event)
                self.notifyListeners(evt)
```

Here's another example used when we also want to report affiliate internet names (but that's probably not relevant here, as this module queries specifically for subdomains):

```python
        for domain in set(domains):
            if self.getTarget().matches(domain, includeChildren=True):
                evt_type = 'INTERNET_NAME'
            else:
                evt_type = 'AFFILIATE_INTERNET_NAME'

            if self.opts['verify'] and not self.sf.resolveHost(domain):
                self.sf.debug(f"Host {domain} could not be resolved")
                evt_type += '_UNRESOLVED'

            evt = SpiderFootEvent(evt_type, domain, self.__name__, event)
            self.notifyListeners(evt)
```

Also, it is common practice to also report `DOMAIN_NAME` in every module where `INTERNET_NAME` is reported and the `INTERNET_NAME` is also a `DOMAIN_NAME`. This hasn't been applied universally across all modules, but should be. Here's a couple of examples:

```python
        for host in set(hosts):
            evt = SpiderFootEvent('INTERNET_NAME', host, self.__name__, event)
            self.notifyListeners(evt)
            if self.sf.isDomain(host, self.opts['_internettlds']):
                evt = SpiderFootEvent('DOMAIN_NAME', host, self.__name__, event)
                self.notifyListeners(evt)
```

```python
        for domain in set(domains):

            # ...

            if self.sf.isDomain(domain, self.opts['_internettlds']):
                if evt_type.startswith('AFFILIATE'):
                    evt = SpiderFootEvent('AFFILIATE_DOMAIN_NAME', domain, self.__name__, event)
                    self.notifyListeners(evt)
                else:
                    evt = SpiderFootEvent('DOMAIN_NAME', domain, self.__name__, event)
                    self.notifyListeners(evt)
```

leotrubach:
Thanks, will do

PR #1004, opened 2020-09-08
Feature/ipregistry.co

leotrubach:


modules/sfp_ipregistry.py
```diff
@@ -0,0 +1,169 @@
...
+# -------------------------------------------------------------------------------
+
+import json
+import urllib.parse
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_ipregistry(SpiderFootPlugin):
+    __name__ = 'sfp_ipregistry'
```
smicallef, line 20: Why is this needed?

modules/sfp_ipregistry.py
```diff
@@ -0,0 +1,169 @@
...
+    meta = {
+        "name": "IP Registry",
```
smicallef, line 22: The name is technically `ipregistry` according to their website so I suggest we go with that. Please update this across the whole file.

modules/sfp_ipregistry.py
```diff
@@ -0,0 +1,169 @@
...
+            self.emit("PHYSICAL_COORDINATES", physical_location, pevent)
+
+    def generate_security_events(self, security, pevent):
+        if not isinstance(security, dict):
+            return
+        malicious = any(
+            security.get(k) for k in ("is_abuser", "is_attacker", "is_threat")
+        )
+        if malicious:
+            self.emit("MALICIOUS_IP_ADDRESS", pevent.data, pevent)
```
smicallef, line 141: This should be formatted the same as other modules reporting maliciousness.

modules/sfp_ipregistry.py
leotrubach: If __name__ is not set the following code
```
self.sf.error(f"You enabled {self.__name__} but did not set an API key!", False)
```
outputs ugly message.

bcoles: Try this.

```python
        if self.opts["api_key"] == "":
            self.sf.error(
                f"You enabled {self.__class__.__name__} but did not set an API key!",
                False,
            )
            self.errorState = True
            return None
```

bcoles: Alternatively, hardcode the module name.

```python
        if self.opts["api_key"] == "":
            self.sf.error(
                "You enabled sfp_ipregistry but did not set an API key!",
                False,
            )
            self.errorState = True
            return None
```

PR #1006, opened 2020-09-09
Implement Bitcoin Who's Who functionality

leotrubach:


modules/sfp_bitcoinwhoswho.py
```diff
@@ -0,0 +1,133 @@
...
+# -------------------------------------------------------------------------------
+
+import json
+import urllib.parse
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_bitcoinwhoswho(SpiderFootPlugin):
+    __name__ = 'sfp_bitcoinwhoswho'
```
smicallef, line 20: Not sure why this is needed.

modules/sfp_bitcoinwhoswho.py
```diff
@@ -0,0 +1,133 @@
...
+    meta = {
+        'name': "Bitcoin Who's Who",
+        'summary': "Check for malicious Bitcoin address agains BitcoinWhosWho database",
```
smicallef, line 23: Typo (agains -> against) and we should be consistent with the name - is it BitcoinWhosWho or Bitcoin Who's Who?

modules/sfp_bitcoinwhoswho.py
```diff
@@ -0,0 +1,133 @@
...
+            self.notifyListeners(evt)
+        return evt
+
+    def generate_events(self, data, pevent):
+        if not isinstance(data, dict):
+            return
+
+        scams = data.get("scams", [])
+        if scams:
+            self.emit("MALICIOUS_BITCOIN_ADDRESS", pevent.data)
```
smicallef, line 109: Same as with other `MALICIOUS_` types, the content should be consistent with the others.

PR #1013, opened 2020-09-09
Propose updates to template

leotrubach:


smicallef:
Let's not make any more changes here.. the module architecture will undergo a heavy review soon so I'd rather do that holistically (and update modules) rather than working around the edges.

PR #1023, opened 2020-09-11
Verify hostnames

leotrubach:


modules/sfp_zetalytics.py
```diff
@@ -60,14 +62,21 @@ def watchedEvents(self):
...
             self.notifyListeners(evt)
         return evt
 
+    def verify_emit_internet_name(self, hostname, pevent):
+        if self.opts["verify"] and not self.sf.resolveHost(hostname):
+            self.sf.debug(f"Host {hostname} could not be resolved")
+            self.emit("INTERNET_NAME_UNRESOLVED", hostname, pevent)
+        else:
+            self.emit("INTERNET_NAME", hostname, pevent)
+
```
bcoles, line 79: Rather than duplicate the `emit` function with a `verify_emit_internet_name` variant, I think from a design perspective it makes more sense to have the resolution logic outside of the `emit` function.

leotrubach:
```
        if self.opts["verify"] and not self.sf.resolveHost(hostname):
            self.sf.debug(f"Host {hostname} could not be resolved")
            self.emit("INTERNET_NAME_UNRESOLVED", hostname, pevent)
        else:
            self.emit("INTERNET_NAME", hostname, pevent)
```
This logic is used twice in functions below. Do you suggest to have 2 copies of it instead of shorter `verify_emit_internet_name` call?

bcoles: > This logic is used twice in functions below. Do you suggest to have 2 copies of it instead of shorter `verify_emit_internet_name` call?

Yes, unless you can think of a good reason to include it. I'm ok with a small amount of code duplication if it makes the module more intuitive.

However, this module is [missing some code patterns for determining whether the identified data is within the target scope](https://github.com/smicallef/spiderfoot/pull/992#issuecomment-690991443) using `self.getTarget()`. Adding in these code patterns may increase the code duplication sufficiently such that this helper function is warranted.

leotrubach:
@bcoles could you please check and tell if all is good. Could you also elaborate this part:

`All INTERNET_NAME events are target children of a DOMAIN_NAME event. Presumably a query for subdomains on the DOMAIN_NAME should identify all INTERNET_NAME subdomains in one query, rather than querying for every INTERNET_NAME individually, unless there's some sort of response length restrictions.
`

bcoles:
> 
> 
> @bcoles could you please check and tell if all is good. Could you also elaborate this part:
> 
> `All INTERNET_NAME events are target children of a DOMAIN_NAME event. Presumably a query for subdomains on the DOMAIN_NAME should identify all INTERNET_NAME subdomains in one query, rather than querying for every INTERNET_NAME individually, unless there's some sort of response length restrictions. `

Currently, the code searches for subdomains of every `INTERNET_NAME`:

```python
        if eventName == "INTERNET_NAME":
            data = self.query_subdomains(eventData)
            self.generate_subdomains_events(data, event)

            data = self.query_hostname(eventData)
            self.generate_hostname_events(data, event)

        elif eventName == "DOMAIN_NAME":
            data = self.query_email_domain(eventData)
            self.generate_email_domain_events(data, event)
```

This results in hundreds or thousands of queries to retrieve subdomains. Presumably, every subdomain could also be retrieved by instead searching for subdomains of the `DOMAIN_NAME` once.

I suggest this:

```python
        if eventName == "INTERNET_NAME":
            data = self.query_hostname(eventData)
            self.generate_hostname_events(data, event)

        elif eventName == "DOMAIN_NAME":
            data = self.query_subdomains(eventData)
            self.generate_subdomains_events(data, event)

            data = self.query_email_domain(eventData)
            self.generate_email_domain_events(data, event)
```

modules/sfp_zetalytics.py
```diff
@@ -60,13 +63,28 @@ def watchedEvents(self):
         return ["INTERNET_NAME", "DOMAIN_NAME", "EMAILADDR"]
 
     def producedEvents(self):
-        return ["INTERNET_NAME", "AFFILIATE_DOMAIN_NAME"]
+        return ["INTERNET_NAME", "AFFILIATE_DOMAIN_NAME", "INTERNET_NAME_UNRESOLVED"]
 
     def emit(self, etype, data, pevent):
+        if self.checkForStop():
+            raise ModuleStop()
```
smicallef, line 70: Please stick with the existing approach for now. We will be revisiting the approach to this later on.

PR #1041, opened 2020-09-13
Add sfp_dnsdb module and test

faleksic:


modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+import json
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+import time
+
+
+class sfp_dnsdb(SpiderFootPlugin):
+    meta = {
+        "name": "DNSDB",
+        "summary": "Resolve and get history of some domain and IP",
```
smicallef, line 22: Query FarSight's DNSDB for historical and passive DNS data.

modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+
+        try:
+            records = []
+            for content in splittedContent:
+                records.append(json.loads(content))
+        except Exception as e:
+            self.sf.error(f"Error processing JSON response from DNSDB: {e}", False)
+            return None
+
+        return records[1:-1]
```
smicallef, line 135: Any likelihood of an out of bounds error here if the length of `splittedContent` isn't as expected?

modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+            rrsetRecords = self.query("rrset", "name", eventData)
+            if rrsetRecords is None:
+                return None
+
+            for record in rrsetRecords:
+                record = record.get("obj")
+                if self.checkForStop():
+                    return None
+
+                if self.isTooOld(record.get("time_last")):
```
smicallef, line 183: Any chance `.get()` returns `None` here? Might make sense to make this `.get("time_last", 0)` instead.

modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+                    "A",
+                    "AAAA",
+                    "MX",
+                    "NS",
+                    "TXT",
+                    "CNAME",
+                ):
+                    continue
+
+                self.emit("RAW_RIR_DATA", str(record), event)
```
smicallef, line 196: `RAW_RIR_DATA` should only be sent once, representing the full contents of the API call, not for every record returned from a single API call.

modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+                if record.get("rrtype") not in ("A", "AAAA"):
+                    continue
+
+                data = record.get("rrname").rstrip(".")
+
+                if data in responseData:
+                    continue
+                responseData.add(data)
+
+                self.emit("RAW_RIR_DATA", str(record), event)
```
smicallef, line 299: See previous `RAW_RIR_DATA` comment.

modules/sfp_dnsdb.py
```diff
@@ -0,0 +1,333 @@
...
+                    "NS",
+                    "TXT",
+                    "CNAME",
+                ):
+                    continue
+
+                self.emit("RAW_RIR_DATA", str(record), event)
+                for data in record.get("rdata"):
+                    data = data.rstrip(".")
+                    if data in responseData:
```
smicallef, line 199: Does the whole data need to be tracked? Maybe just a hash?

modules/sfp_dnsdb.py
faleksic: I fixed this by checking is`splittedContent` less or equal to two, if it is it will return None, else it will return records

PR #1085, opened 2020-10-02
Debounce module

krishnasism:


krishnasism:
Ignore the _fetchtimeoutfix commit, it didn't fix it.

test/unit/modules/test_sfp_debounce.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_watchedEvents_should_return_list(self):
+        module = sfp_debounce()
+        self.assertIsInstance(module.watchedEvents(), list)
+
+    def test_producedEvents_should_return_list(self):
+        module = sfp_debounce()
+        self.assertIsInstance(module.producedEvents(), list)
+
+    def test_handleEvent(self):
```
bcoles, line 57:
```suggestion
    @unittest.skip("todo")
    def test_handleEvent(self):
```

The tests are failing due to the code pattern used in the unit test.

```
test/unit/modules/test_sfp_debounce.py:77: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

modules/sfp_debounce.py:99: in handleEvent

    data = self.queryEmailAddr(eventData)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <modules.sfp_debounce.sfp_debounce object at 0x7f6df8370700>

qry = 'example data'

    def queryEmailAddr(self, qry):

        res = self.sf.fetchUrl(

            f"https://disposable.debounce.io?email={qry}",

>           timeout=self.opts['_fetchtimeout'],

            useragent="SpiderFoot"

        )

E       KeyError: '_fetchtimeout'

modules/sfp_debounce.py:71: KeyError
```

The easiest way to fix this, for now, is to mark the test to be skipped.

bcoles:
Here's the config I use for tests. Testing locally is faster than waiting to see if Travis fails.

```
$ cat ./test/run
#!/bin/sh
rm .pytest_cache/ -rf
python3 -m pytest --flake8 -n 2 --dist loadfile --durations=5 --cov-report html --cov=. .
```

The tests take around 3 minutes on a VM with 2 CPU cores.

Removing the coverage reporting `--cov-report html --cov=. .` would also increase the speed of the tests.

test/unit/modules/test_sfp_debounce.py
krishnasism: @bcoles thanks! It removed the '_fetchtimeout' error.

modules/sfp_debounce.py
```diff
@@ -0,0 +1,112 @@
...
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        self.results[eventData] = True
+
+        data = self.queryEmailAddr(eventData)
+
+        if data is None:
+            return
+
+        evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
```
smicallef, line 104: I suggest moving this to within the following `if` statement since we only want `RAW_RIR_DATA` if there's content in the API response that has been of value.

spiderfoot/db.py
```diff
@@ -144,6 +144,7 @@ class SpiderFootDb:
         ['DOMAIN_WHOIS', 'Domain Whois', 1, 'DATA'],
         ['EMAILADDR', 'Email Address', 0, 'ENTITY'],
         ['EMAILADDR_COMPROMISED', 'Hacked Email Address', 0, 'DESCRIPTOR'],
+        ['EMAILADDR_DISPOSABLE', 'Disposable Email Address', 0, 'ENTITY'],
```
smicallef, line 147: Should be `DESCRIPTOR` since the disposable state of the email is describing the `EMAILADDR` entity.

PR #1091, opened 2020-10-03
Trumail module

krishnasism:


modules/sfp_trumail.py
```diff
@@ -0,0 +1,115 @@
...
+import json
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_trumail(SpiderFootPlugin):
+
+    meta = {
+        'name': "Trumail",
+        'summary': "Check whether an email is disposable",
+        'flags': ["apikey"],
```
smicallef, line 23: Doesn't seem to need an API key?

PR #1105, opened 2020-10-09
IP Quality Score changes

krishnasism:


modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
         recentAbuse = data.get('recent_abuse')
         botStatus = data.get('bot_status')
+        malicious = False
+        maliciousDesc = ""
+
         if fraudScore >= self.opts['abuse_score_threshold'] or recentAbuse or botStatus:
             evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
             self.notifyListeners(evt)
+            malicious = True
+            maliciousDesc = f"IPQualityScore [ {eventData} ]\n"
```
smicallef, line 170: Should be no spaces between the `[` `]` and `eventData`.

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
+            if city:
+                geoInfo += city + ", "
+            if region:
+                geoInfo += region + ", "
+            if country:
+                geoInfo += country + " "
+            if zipcode:
+                geoInfo += zipcode + ". "
+
+            evt = SpiderFootEvent("GEOINFO", geoInfo, self.__name__, event)
```
smicallef, line 193: `GEOINFO` isn't declared in `producedEvents()`

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
+            region = data.get('region')
+
+            if city:
+                geoInfo += city + ", "
+            if region:
+                geoInfo += region + ", "
+            if country:
+                geoInfo += country + " "
+            if zipcode:
+                geoInfo += zipcode + ". "
```
smicallef, line 191: I don't think you need a `.`.

modules/sfp_ipqualityscore.py
smicallef, line 193: `geoinfo` may still be empty by this stage, so you should wrap this in a `if geoInfo` clause.

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
-                evt = SpiderFootEvent("MALICIOUS_EMAILADDR", "IPQualityScore [" + eventData + "]", self.__name__, event)
+
+            if data.get('disposable'):
+                evt = SpiderFootEvent("EMAILADDR_DISPOSABLE", eventData, self.__name__, event)
                 self.notifyListeners(evt)
-            elif eventName == "IP_ADDRESS":
-                evt = SpiderFootEvent("MALICIOUS_IPADDR", "IPQualityScore [" + eventData + "]", self.__name__, event)
+
+            if data.get('leaked'):
+                evt = SpiderFootEvent("EMAILADDR_COMPROMISED", f"{eventData} [ {data.get('suggested_domain')} ]", self.__name__, event)
```
smicallef, line 207: Please remove paces after `[` and before `]`.

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
+
+            if city:
+                geoInfo += city + ", "
+            if region:
+                geoInfo += region + ", "
+            if country:
+                geoInfo += country + " "
+
+            evt = SpiderFootEvent("GEOINFO", geoInfo, self.__name__, event)
+            self.notifyListeners(evt)
```
smicallef, line 231: Since this code for geo data is duplicated, maybe move it to its own function?

smicallef, line 231: 3045835950

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +160,74 @@ def handleEvent(self, event):
...
+        malicious = False
+        maliciousDesc = ""
+
         if fraudScore >= self.opts['abuse_score_threshold'] or recentAbuse or botStatus:
             evt = SpiderFootEvent("RAW_RIR_DATA", str(data), self.__name__, event)
             self.notifyListeners(evt)
+            malicious = True
+            maliciousDesc = f"IPQualityScore [ {eventData} ]\n"
+
+        if eventName == "PHONE_NUMBER":
```
smicallef, line 172: I believe the API also indicates the phone number type, or at least if it's disposable? That way this module could return `PHONE_NUMBER_TYPE`, and we could also add `PHONE_NUMBER_DISPOSABLE` as a type, just as we have for emails.

krishnasism: Don't see anything like this in the response

modules/sfp_ipqualityscore.py
```diff
@@ -156,21 +183,58 @@ def handleEvent(self, event):
...
-                evt = SpiderFootEvent("MALICIOUS_EMAILADDR", "IPQualityScore [" + eventData + "]", self.__name__, event)
+
+            if data.get('disposable'):
+                evt = SpiderFootEvent("EMAILADDR_DISPOSABLE", eventData, self.__name__, event)
                 self.notifyListeners(evt)
-            elif eventName == "IP_ADDRESS":
-                evt = SpiderFootEvent("MALICIOUS_IPADDR", "IPQualityScore [" + eventData + "]", self.__name__, event)
+
+            if data.get('leaked'):
+                evt = SpiderFootEvent("EMAILADDR_COMPROMISED", f"{eventData} [{data.get('suggested_domain')}]", self.__name__, event)
```
smicallef, line 222: Why `suggested_domain`?

PR #1108, opened 2020-10-13
Use Bootstrap form-group instead of input-group

bairdj:
Each option field only has one form control, so there is no need to use `input-group`. Instead, `form-group` can be used. This PR also adds Bootstrap styling to `select` elements


smicallef:
Thanks for the contribution!

PR #1114, opened 2020-10-25
sfp_leakix: Added api-key support

gboddin:
Hi, I'm adding this before tomorrow's deployment so it's ready just in case !

I'm not sure I should add `apikey` in `tags` since it won't be strictly required. The results will be limited in time if it's not present though.

Thanks for the orignal plugin @bcoles and the whole team for the tool :heart:

### DIFF
--- a/modules/sfp_leakix.py
+++ b/modules/sfp_leakix.py
@@ -26,7 +26,7 @@ class sfp_leakix(SpiderFootPlugin):
         'categories': ["Leaks, Dumps and Breaches"],
         'dataSource': {
             'website': "https://leakix.net/",
-            'model': "FREE_NOAUTH_UNLIMITED",
+            'model': "FREE_AUTH_UNLIMITED",
             'references': [
                 "https://leakix.net/api-documentation"
             ],
@@ -40,6 +40,7 @@ class sfp_leakix(SpiderFootPlugin):
 
     # Default options
     opts = {
+        'api_key': "",
         'delay': 1,
     }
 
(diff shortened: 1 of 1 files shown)


bcoles (changes requested):
Thanks for this.

The changes look good but the module is missing a description for the new `api_key` option which is causing the tests to fail.

```
            # check len(options) == len(option descriptions)

            if m.get('opts'):

>               self.assertEqual("%s opts: %s" % (module, len(m.get('opts').keys())), "%s opts: %s" % (module, len(m.get('optdescs').keys())))

E               AssertionError: 'sfp_leakix opts: 2' != 'sfp_leakix opts: 1'

E               - sfp_leakix opts: 2

E               ?                  ^

E               + sfp_leakix opts: 1

E               ?                  ^

test/unit/test_spiderfoot_module_loading.py:104: AssertionError

__________________________ TestModuleleakix.test_opts __________________________

self = <test.unit.modules.test_sfp_leakix.TestModuleleakix testMethod=test_opts>

    def test_opts(self):

        module = sfp_leakix()

>       self.assertEqual(len(module.opts), len(module.optdescs))

E       AssertionError: 2 != 1

test/unit/modules/test_sfp_leakix.py:38: AssertionError
```

This can easily be resolved by adding `"api_key": "LeakIX API key.",` to the `optdescs` dict.

The `meta` info dict is also missing instructions on signing up for an API key.

gboddin:
Thank you, fixed !

PR #1120, opened 2020-11-25
Added docker compose file

swedishmike:
Just added a simple docker-compose.yml file for those who like to use that as well.


smicallef:
Great, thanks for the contribution!

PR #1136, opened 2021-01-25
Grayhat Warfare module

krishnasism:


modules/sfp_grayhatwarfare.py
```diff
@@ -0,0 +1,169 @@
...
+
+        headers = {
+            'Accept': 'application/json',
+        }
+
+        res = self.sf.fetchUrl(queryString,
+                               headers=headers,
+                               timeout=15,
+                               useragent=self.opts['_useragent'],
+                               verify=True)
```
bcoles, line 91: The usual code pattern is to `import urllib` and use `urllib.parse.urlencode` to prevent issues with encoding.

```suggestion
        params = urllib.parse.urlencode({
            'keywords': keyword.encode('raw_unicode_escape'),
            'access_token': self.opts['api_key']
        })

        headers = {
            'Accept': 'application/json',
        }

        res = self.sf.fetchUrl(
            f"https://buckets.grayhatwarfare.com/api/v1/buckets/{start}/{self.opts['per_page']}?{params}",
            headers=headers,
            timeout=15,
            useragent=self.opts['_useragent'],
            verify=True
        )
```

PR #1137, opened 2021-01-29
Etherscan module

krishnasism:


modules/sfp_etherscan.py
```diff
@@ -0,0 +1,131 @@
...
+    def handleEvent(self, event):
+        eventName = event.eventType
+        srcModuleName = event.module
+        eventData = event.data
+
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        if self.opts['api_key'] == "":
+            self.sf.error("You enabled sfp_etherscan but did not set an API key!")
+            self.errorState = True
```
smicallef, line 112: This module doesn't appear to be checking if errorState is True anywhere?

modules/sfp_etherscan.py
```diff
@@ -0,0 +1,131 @@
...
+            return
+
+        # Don't look up stuff twice
+        if eventData in self.results:
+            self.sf.debug(f"Skipping {eventData}, already checked.")
+            return None
+        else:
+            self.results[eventData] = True
+
+        data = self.query(eventData)
```
smicallef, line 122: What if `query` returns an empty value or error?

modules/sfp_etherscan.py
```diff
@@ -0,0 +1,131 @@
...
+        balance = float(data.get('result')) / 1000000000000000000
```
smicallef, line 123: A comment explaining this would be useful.

PR #1138, opened 2021-01-31
ipapi.com module

krishnasism:


modules/sfp_ipapicom.py
```diff
@@ -0,0 +1,132 @@
...
+            "RAW_RIR_DATA"
+        ]
+
+    def query(self, qry):
+        queryString = f"http://api.ipapi.com/api/{qry}?access_key={self.opts['api_key']}"
+
+        res = self.sf.fetchUrl(queryString,
+                               timeout=self.opts['_fetchtimeout'],
+                               useragent=self.opts['_useragent'])
+
```
bcoles, line 90:
```suggestion
        time.sleep(1.5)

        if res['code'] == "429":
            self.sf.error("You are being rate-limited by IP-API.com.")
            self.errorState = True
            return None
```

modules/sfp_ipapicom.py
```diff
@@ -0,0 +1,137 @@
...
+    def handleEvent(self, event):
+        eventName = event.eventType
+        srcModuleName = event.module
+        eventData = event.data
+
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        if self.opts['api_key'] == "":
+            self.sf.error("You enabled sfp_ipapicom but did not set an API key!")
+            self.errorState = True
```
smicallef, line 115: errorState is set but not checked.

modules/sfp_ipapicom.py
```diff
@@ -0,0 +1,140 @@
...
+            self.sf.debug(f"Skipping {eventData}, already checked.")
+            return None
+        else:
+            self.results[eventData] = True
+
+        data = self.query(eventData)
+
+        if data.get('country_name'):
+            location = ', '.join(filter(None, [data.get('city'), data.get('region_name'), data.get('region_code'), data.get('country_name'), data.get('country_code')]))
+            location += f"\n-Latitude: {data.get('latitude')}\n-Longitude: {data.get('longitude')}"
```
smicallef, line 132: Lat/Lon should be reported separately as `PHYSICAL_COORDINATES`.

modules/sfp_ipapicom.py
```diff
@@ -0,0 +1,140 @@
...
+        eventData = event.data
+
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        if self.opts['api_key'] == "":
+            self.sf.error("You enabled sfp_ipapicom but did not set an API key!")
+            self.errorState = True
+            return None
+
+        if self.errorState:
```
smicallef, line 118: This check should be before the above `if` block, since otherwise an error log will be generated for every event received.

PR #1141, opened 2021-02-13
Seon.io module

krishnasism:


modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+    # Default options
+    opts = {
+        'api_key': '',
+        'fraud_threshold': 80,
+    }
+
+    # Option descriptions
+    optdescs = {
+        'api_key': "API Key for seon.io",
+        'fraud_threshold': 'Minimum fraud score for target to be marked as malicious(0-100)',
```
smicallef, line 49: Missing space after malicious.

modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+    # This is to support the end user in selecting modules based on events
+    # produced.
+    def producedEvents(self):
+        return [
+            "GEOINFO",
+            "MALICIOUS_IPADDR",
+            "TCP_PORT_OPEN",
+            "MALICIOUS_EMAILADDR",
+            "EMAILADDR_DELIVERABLE",
+            "EMAILADDR_UNDELIVERABLE",
```
smicallef, line 80: Typo

modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+            "SOCIAL_MEDIAL",
```
smicallef, line 81: Typo

modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+                    location += f"\n-Latitude: {resultSet.get('latitude')}\n-Longitude: {resultSet.get('longitude')}"
+                    evt = SpiderFootEvent('GEOINFO', location, self.__name__, event)
+                    self.notifyListeners(evt)
+
+                if resultSet.get('open_ports'):
+                    for port in resultSet.get('open_ports'):
+                        evt = SpiderFootEvent('TCP_PORT_OPEN', f"{eventData}:{port}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                if resultSet.get('tor'):
```
smicallef, line 160: These following events should only be reported if they are positive. i.e. I don't think it's valuable to report that something is NOT a VPN, for example, since most things won't be.

modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+                    if resultSet.get('domain_details').get('disposable'):
+                        evt = SpiderFootEvent("EMAILADDR_DISPOSABLE", eventData, self.__name__, event)
+                        self.notifyListeners(evt)
+
+                if resultSet.get('account_details'):
+                    socialMediaList = resultSet.get('account_details').keys()
+                    for site in socialMediaList:
+                        if resultSet.get('account_details').get(site):
+                            if resultSet.get('account_details').get(site).get('url'):
+                                evt = SpiderFootEvent("SOCIAL_MEDIA", f"{site}: <SFURL> {resultSet.get('account_details').get(site).get('url')} </SFURL>", self.__name__, event)
```
smicallef, line 205: No space after <SFURL> and before </SFURL>.

modules/sfp_seon.py
```diff
@@ -0,0 +1,256 @@
...
+
+                if resultSet.get('type'):
+                    evt = SpiderFootEvent("PHONE_NUMBER_TYPE", resultSet.get('type'), self.__name__, event)
+                    self.notifyListeners(evt)
+
+                if resultSet.get('carrier'):
+                    evt = SpiderFootEvent("PROVIDER_TELCO", resultSet.get('carrier'), self.__name__, event)
+                    self.notifyListeners(evt)
+
+            evt = SpiderFootEvent('RAW_RIR_DATA', str(resultSet), self.__name__, event)
```
smicallef, line 253: `RAW_RIR_DATA` should only be reported if data was found.

modules/sfp_seon.py
```diff
@@ -0,0 +1,262 @@
...
+
+    # Default options
+    opts = {
+        'api_key': '',
+        'fraud_threshold': 80,
+    }
+
+    # Option descriptions
+    optdescs = {
+        'api_key': "API Key for seon.io",
```
smicallef, line 48: This should be in the form of `<name> API Key.`

modules/sfp_seon.py
```diff
@@ -0,0 +1,262 @@
...
+        eventData = event.data
+
+        self.sf.debug(f"Received event, {eventName}, from {srcModuleName}")
+
+        if self.opts['api_key'] == "":
+            self.sf.error("You enabled sfp_seon but did not set an API key!")
+            self.errorState = True
+            return
+
+        if self.errorState:
```
smicallef, line 127: Check should be higher up.

modules/sfp_seon.py
```diff
@@ -0,0 +1,262 @@
...
+                        evt = SpiderFootEvent("WEBSERVER_TECHNOLOGY", f"Server is Web Proxy: {resultSet.get('web_proxy')}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                    if resultSet.get('public_proxy'):
+                        evt = SpiderFootEvent("WEBSERVER_TECHNOLOGY", f"Server is Public Proxy: {resultSet.get('public_proxy')}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                if resultSet.get('country'):
+                    location = ', '.join(filter(None, [resultSet.get('city'), resultSet.get('state_prov'), resultSet.get('country')]))
+                    location += f"\n-Latitude: {resultSet.get('latitude')}\n-Longitude: {resultSet.get('longitude')}"
```
smicallef, line 167: Lat/Lon has its own event type - `PHYSICAL_COORDINATES`, so should be a separate additional event.

modules/sfp_seon.py
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        res = self.sf.fetchUrl(
+            queryString,
+            headers=headers,
+            timeout=15,
+            useragent=self.opts['_useragent']
+        )
+
+        return json.loads(res['content'])
```
smicallef, line 112: Should have the regular 404, 429, etc. checks here before returning and setting `errorState` if appropriate.

modules/sfp_seon.py
```diff
@@ -0,0 +1,263 @@
...
+            self.results[eventData] = True
+
+        if eventName == "IP_ADDRESS" or eventName == "IPV6_ADDRESS":
+            data = self.query(eventData, eventName)
+            if data is None:
+                return None
+
+            resultSet = data.get('data')
+            if resultSet:
+                if resultSet.get('score') >= self.opts['fraud_threshold']:
```
smicallef, line 144: When using `.get()` and expecting an integer for comparison, set `0` as the value to be returned in case the key isn't found, so that you don't raise an exception for trying to compare a `None` with a number.

modules/sfp_seon.py
```diff
@@ -0,0 +1,263 @@
...
+                        evt = SpiderFootEvent("WEBSERVER_TECHNOLOGY", f"Server is VPN: {resultSet.get('vpn')}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                    if resultSet.get('web_proxy'):
+                        evt = SpiderFootEvent("WEBSERVER_TECHNOLOGY", f"Server is Web Proxy: {resultSet.get('web_proxy')}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                    if resultSet.get('public_proxy'):
+                        evt = SpiderFootEvent("WEBSERVER_TECHNOLOGY", f"Server is Public Proxy: {resultSet.get('public_proxy')}", self.__name__, event)
+                        self.notifyListeners(evt)
```
smicallef, line 163: If the entity isn't a proxy, etc. will this still result in an event? It should not - i.e. we only want events here if the entity is a proxy, etc.

modules/sfp_seon.py
```diff
@@ -0,0 +1,263 @@
...
+                    self.notifyListeners(evt)
+
+                    evt = SpiderFootEvent('PHYSICAL_COORDINATES', f"{resultSet.get('latitude')}, {resultSet.get('longitude')}", self.__name__, event)
+                    self.notifyListeners(evt)
+                if resultSet.get('open_ports'):
+                    for port in resultSet.get('open_ports'):
+                        evt = SpiderFootEvent('TCP_PORT_OPEN', f"{eventData}:{port}", self.__name__, event)
+                        self.notifyListeners(evt)
+
+                evt = SpiderFootEvent('RAW_RIR_DATA', str(resultSet), self.__name__, event)
```
smicallef, line 177: `RAW_RIR_DATA` should only be reported if data was found for the entity.

modules/sfp_seon.py
krishnasism: It should not go in if its False

PR #1142, opened 2021-02-26
Social links module - mtg-bi.com

krishnasism:


modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,290 @@
...
+        'useCases': ["Footprint", "Investigate", "Passive"],
+        'categories': ["Real World"],
+        'dataSource': {
+            'website': "https://mtg-bi.com/",
+            'model': "COMMERCIAL_ONLY",
+            'references': [
+                "https://docs.osint.rest/"
+            ],
+            'favIcon': "https://seon.io/assets/favicons/favicon-16x16.png",
+            'logo': "https://seon.io/assets/favicons/apple-touch-icon-152.png",
```
smicallef, line 34: Copy-paste error :)

modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    # Default options
+    opts = {
+        'api_key': '',
+    }
+
+    # Option descriptions
+    optdescs = {
+        'api_key': "API Key for mtg-bi.com",
```
smicallef, line 47: Convention is "<name> API Key."

modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,290 @@
...
+    def queryLinkedin(self, qry, eventName):
+        if eventName == "EMAILADDR":
+            queryString = f"https://osint.rest/api/linkedin/lookup_by_email/v2?query={qry}"
+
+        headers = {
+            'Accept': "application/json",
+            'Authorization': self.opts['api_key']
+        }
+
+        res = self.sf.fetchUrl(
```
smicallef, line 151: There is a lot of code duplication in these query* functions. I suggest creating a generic query function and moving the common code there, then having the query* functions call that.

modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,290 @@
...
+            if data is None:
+                failedModules += 1
+            else:
+                resultSet = data.get('result')
+                if resultSet:
+                    resultSet = data.get('result')[0].get('nodeProfileData')
+                    if resultSet.get('name'):
+                        humanNames.add(resultSet.get('name'))
+                    if resultSet.get('skypeId'):
+                        evt = SpiderFootEvent("USERNAME", resultSet.get('skypeId'), self.__name__, event)
```
smicallef, line 251: This should be an `ACCOUNT_EXTERNAL_OWNED` event instead - see the sfp_c99 module for how it reports Skype accounts.

modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,254 @@
...
+        res = self.sf.fetchUrl(
+            queryString,
+            headers=headers,
+            timeout=15,
+            useragent=self.opts['_useragent']
+        )
+
+        if res['content'] is None:
+            return None
+        return json.loads(res['content'])
```
smicallef, line 97: Remember to check for error codes.

modules/sfp_sociallinks.py
```diff
@@ -0,0 +1,254 @@
...
+    def queryTelegram(self, qry, eventName):
+        if eventName == "PHONE_NUMBER":
+            queryString = f"https://osint.rest/api/telegram/user_by_phone?query={qry}"
+        elif eventName == "USERNAME":
+            queryString = f"https://osint.rest/api/telegram/user_by_alias?query={qry}"
+
+        return self.query(queryString)
+
+    def queryFlickr(self, qry, eventName):
+        if eventName == "EMAILADDR":
```
smicallef, line 108: What is the purpose of these various query* calls which have a single `if` statement checking the `eventName`? If the function only takes one kind of event type, shouldn't the caller ensure that's what it's called with?

PR #1143, opened 2021-03-01
changed instructions to 3.3

ZephrFish:


smicallef:
There's no real benefit to this change, since when pulled form master or taken from the release tarball, the folder is always `spiderfoot`.

PR #1155, opened 2021-04-18
HaveIBeenPwned - Paste site update

krishnasism:


modules/sfp_haveibeenpwned.py
```diff
@@ -135,23 +179,54 @@ def handleEvent(self, event):
...
+                try:
+                    if not self.opts['api_key']:
+                        site = n["Title"]
+                    else:
+                        site = n["Name"]
+                except Exception as e:
+                    self.sf.debug(f"Unable to parse result from HaveIBeenPwned?: {e}")
+                    continue
+
+                evt = eventName + "_COMPROMISED"
```
smicallef, line 193: `eventName` will always be `EMAILADDR`, right? Better to be explicit, if that's the case.

modules/sfp_haveibeenpwned.py
```diff
@@ -135,23 +179,54 @@ def handleEvent(self, event):
             self.results[eventData] = True
 
         data = self.query(eventData)
-        if data is None:
+        if data is not None:
+            for n in data:
+                try:
+                    if not self.opts['api_key']:
```
smicallef, line 185: Different fields depending on whether an API key is used or not? Isn't an API key always required for HIBP?

modules/sfp_haveibeenpwned.py
```diff
@@ -135,23 +179,54 @@ def handleEvent(self, event):
...
+                site = source
+                if source in sites.keys():
+                    site = sites[n.get("Source")]
+
+                if not n.get("Source") is None:
+                    e = SpiderFootEvent("LEAKSITE_URL", site, self.__name__, event)
+                    self.notifyListeners(e)
+
+                if not n.get("Title") is None:
+                    e = SpiderFootEvent("LEAKSITE_CONTENT", n.get("Title"), self.__name__, event)
```
smicallef, line 224: Look at sfp_pastebin to see how `LEAKSITE_CONTENT` should be reported.

modules/sfp_haveibeenpwned.py
```diff
@@ -135,23 +179,54 @@ def handleEvent(self, event):
...
-                    site = n["Title"]
-                else:
-                    site = n["Name"]
+                source = n.get("Source")
+                site = source
+                if source in sites.keys():
+                    site = sites[n.get("Source")]
+
+                if not n.get("Source") is None:
+                    e = SpiderFootEvent("LEAKSITE_URL", site, self.__name__, event)
```
smicallef, line 220: Won't `site`' just be the URL of the paste site? It should be the URL of the actual paste itself on that paste site.

modules/sfp_haveibeenpwned.py
krishnasism: This was there from the beginning, maybe some different logic back then?

modules/sfp_haveibeenpwned.py
krishnasism: This is all we receive: 

{
"Source":"Pastebin",
"Id":"8Q0BvKD8",
"Title":"syslog",
"Date":"2014-03-04T19:14:54Z",
"EmailCount":139
},

modules/sfp_haveibeenpwned.py
krishnasism: The whole of res['content'] is being reported, in this case should it be str(pasteData) ?

PR #1168, opened 2021-05-19
DNS Brute Module Rewrite

TheTechromancer:
This is a rewrite of the DNSBrute module with the following changes/enhancements:
- Automatic round-robin across many DNS servers to avoid rate-limiting
- Automatic validation of every nameserver and every newly discovered subdomain
- Automatic tolerance against network errors, DNS server errors, or garbage data
- Smart alpha mutations such as "-old", "-test", "-prod", as well as sub-variations of existing subdomains
- Smart number mutations which increment/decrement existing numbers in addition to appending 1, 01, 001, etc.
- Numeric, alpha, and subdomain-of-subdomain brute-forcing is now applied to `INTERNET_NAME_UNRESOLVED` as well as `INTERNET_NAME`

The new logic goes like this:
1. During module setup, validate each public DNS server in `dicts/resolvers.txt` by making a simple query and discarding the server if either the result is invalid or if it doesn't respond within 2 seconds
2. When an event is triggered, iterate through all the applicable brute-force modes, load-balancing across the valid resolvers established in step 1 (NOTE: If there is an error with the DNS server, it will try again using a different one, up to ten times.)
3. Validate each discovered record against known-good DNS servers (`8.8.8.8` and `1.1.1.1`) before triggering an event

Note that this module uses a new `ThreadPool` class in `spiderfoot/plugin.py` which integrates with `SpiderFootPlugin` for simpler threading within modules. It uses simple threads and a queue to achieve this, and constantly uses `checkForStop()`  to gracefully stop threads. Below is an example of how to use it (also included in the class docstring):
```
with self.threadPool(self.opts["_maxthreads"]) as pool:
    for result in pool.map(
            ['a', 'b', 'c', 'd'],
            self.functionToThread
        ):
        yield result
```


modules/sfp_dnsbrute.py
```diff
@@ -4,17 +4,17 @@
 # Purpose:      SpiderFoot plug-in for attempting to resolve through brute-forcing
 #               common hostnames.
 #
-# Author:      Steve Micallef <steve@binarypool.com>
+# Author:      TheTechromancer
 #
-# Created:     06/07/2017
-# Copyright:   (c) Steve Micallef 2017
```
smicallef, line 10: Please don't remove the copyright :)

modules/sfp_dnsbrute.py
```diff
@@ -4,17 +4,17 @@
 # Purpose:      SpiderFoot plug-in for attempting to resolve through brute-forcing
 #               common hostnames.
 #
-# Author:      Steve Micallef <steve@binarypool.com>
+# Author:      TheTechromancer
```
smicallef, line 7: Understand this is a substantial re-write, but would prefer you add yourself and comment that it was a substantial re-write.

modules/sfp_dnsbrute.py
```diff
@@ -30,212 +30,396 @@ class sfp_dnsbrute(SpiderFootPlugin):
...
+        "alphamutation": "For any host found, try common mutations such as -test, -old, etc.",
         "_maxthreads": "Maximum threads"
     }
 
-    events = None
-    sublist = None
-    lock = None
-
     def setup(self, sfc, userOpts=dict()):
+
```
smicallef, line 52: Not a fan of the style having an empty newline after the function definition - could you please re-style to fit with the convention of the rest of the codebase?

TheTechromancer:
Thanks for reviewing, comments have been addressed 👍

dicts/resolvers.txt
```diff
@@ -0,0 +1,986 @@
+103.20.188.35
```
smicallef, line 1: Where is this list from? It should be credited somewhere, and would also be good to know how we can keep it updated over time.

TheTechromancer: I have added the following comment at the top of the file:
```
# Public DNS servers
# Reliability: 95% or higher
# Valid as of 2021-06-28
# Retrieved from https://public-dns.info/nameservers.csv
```

Also, the module will now automatically retrieve the latest dataset (only servers with 95% or higher reliability), and fall back to the local dictionary if the request fails or if there aren't any results.

TheTechromancer:
FYI the new CICD tests failed but I don't think it's the PR's fault.
```
=========================== short test summary info ============================
FAILED test/unit/modules/test_sfp_opendns.py::TestModuleopendns::test_queryAddr_should_resolve_unblocked_host
FAILED test/unit/modules/test_sfp_quad9.py::TestModulequad9::test_query_should_resolve_unblocked_host
FAILED test/unit/modules/test_sfp_yandexdns.py::TestModuleyandexdns::test_queryAddr_should_resolve_unblocked_host
==== 3 failed, 1427 passed, 135 skipped, 1742 warnings in 61.05s (0:01:01) =====
Error: Process completed with exit code 1.
```

TheTechromancer:
Temporarily closing this PR before adding some improvements.

PR #1169, opened 2021-05-20
Added reversewhois.io module

TheTechromancer:
https://www.reversewhois.io/ is a free search engine which finds whois records by email address or email domain. This module makes a single domain query which returns all domains containing matching email addresses in their whois records.

image


smicallef:
Nice! I don't see any terms on the website so can make an exception for scraping here. I'll do some testing and merge in the coming days if all checks out.

smicallef:
Thanks for the contribution; works great!

PR #1170, opened 2021-05-21
Added Sublist3r PassiveDNS module

TheTechromancer:
The [Sublist3r tool](https://github.com/aboul3la/Sublist3r) hosts its own passive DNS API at https://api.sublist3r.com. This is one of the primary data sources for the tool, which is actively maintained and has over 6,000 stars on GitHub. Since it is technically an undocumented API, this module is careful to clean, deduplicate, and validate the data before raising events.

image

### DIFF
--- a/modules/sfp_sublist3r.py
+++ b/modules/sfp_sublist3r.py
@@ -59,10 +59,10 @@ class sfp_sublist3r(SpiderFootPlugin):
 
         url = f"https://api.sublist3r.com/search.php?domain={domain}"
         ret = []
-        # mirror sublist3r's headers because yes
         res = self.sf.fetchUrl(
             url,
-            useragent="Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36",
+            useragent=self.opts.get("_useragent", "Spiderfoot"),
+            # mirror sublist3r's headers
             headers={
                 "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8",
                 "Accept-Language": "en-US,en;q=0.8",
@@ -103,7 +103,7 @@ class sfp_sublist3r(SpiderFootPlugin):
         self.state["events"].append(eventDataHash)
 
         for hostname in self.query(domain):
-            if hostname.endswith(domain):
+            if hostname.endswith(domain) and not hostname == domain:
                 self.sendEvent(event, hostname)
             else:
                 self.sf.debug(f"Invalid subdomain: {hostname}")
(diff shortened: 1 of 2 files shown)


smicallef:
How reliable is this API? SpiderFoot actually used to have a module for it but we removed it since the API was down quite often.

TheTechromancer:
As a pentester, I use the command-line tool pretty regularly and it's never failed. It may not have "five nines" availability though, since it's hosted by an individual and not an organization. Personally I think the data it returns is valuable enough to justify a module.

modules/sfp_sublist3r.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # skip if we've already processed this event
+        eventDataHash = self.sf.hashstring(event.data)
+        if eventDataHash in self.state["events"]:
+            self.sf.debug(f"Skipping already-processed event, {event.eventType}, from {event.module}")
+            return
+        self.state["events"].append(eventDataHash)
+
+        for hostname in self.query(domain):
+            if hostname.endswith(domain) and not hostname == domain:
```
smicallef, line 101: This should use the `SpiderFootTarget.matches()` method to verify whether it's part of the target scope.

modules/sfp_sublist3r.py
```diff
@@ -0,0 +1,104 @@
...
+    optdescs = {}
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.sf.debug("Setting up sfp_sublist3r")
+        self.state = self.tempStorage()
+        self.state.update({
+            "events": [],
+        })
+        self.__dataSource__ = "DNS"
```
smicallef, line 46: This shouldn't be overridden since the data source is really Sublist3r, not DNS directly.

modules/sfp_sublist3r.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Option descriptions
+    optdescs = {}
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.sf.debug("Setting up sfp_sublist3r")
+        self.state = self.tempStorage()
+        self.state.update({
+            "events": [],
```
smicallef, line 44: I like the approach taken here but for now can you make this consistent with all other modules (using `self.results`)? We'll move to a better state management approach later on.

modules/sfp_sublist3r.py
```diff
@@ -0,0 +1,104 @@
...
+        })
+        self.__dataSource__ = "DNS"
+
+        self.opts.update(userOpts)
+
+    def watchedEvents(self):
+        return ["DOMAIN_NAME"]
```
smicallef, line 51: Should probably watch for `INTERNET_NAME` too since subdomains aren't `DOMAIN_NAME`s.

PR #1171, opened 2021-05-21
Added DNSDumpster module

TheTechromancer:
Scrapes subdomains (in a safe and responsible manner ;) from https://dnsdumpster.com/.

### DIFF
--- a/modules/sfp_dnsdumpster.py
+++ b/modules/sfp_dnsdumpster.py
@@ -1,0 +1,148 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_dnsdumpster
+# Purpose:     SpiderFoot plug-in for subdomain enumeration using
+#              dnsdumpster.com
+#
+# Author:      TheTechromancer
+#
+# Created:     05/21/2021
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import re
+from bs4 import BeautifulSoup
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_dnsdumpster(SpiderFootPlugin):
+
+    meta = {
+        "name": "DNSDumpster",
+        "summary": "Passive subdomain enumeration using HackerTarget's DNSDumpster",
+        "useCases": ["Investigate", "Footprint", "Passive"],
+        "categories": ["Passive DNS"],
+        "dataSource": {
+            "website": "https://dnsdumpster.com/",
+            "model": "FREE_NOAUTH_UNLIMITED",
+            "description": "DNSdumpster.com is a FREE domain research tool that can discover hosts related to a domain.",
+        }
+    }
+
+    # Default options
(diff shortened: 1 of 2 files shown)


smicallef:
Do you know if the data returned is any different from the HackerTarget API (since this project is run by them)? Unless there's a significant data difference, I'd prefer to rely solely on the HackerTarget API since there's no reliance on scraping there. Generally we are trying to avoid any scraping-based modules.

TheTechromancer:
Yes the data is different. At the time of writing I checked `blacklanternsecurity.com` and dnsdumpster returned entries which were not listed by their API. DNSDumpster may be aggregating from multiple sources.

modules/sfp_dnsdumpster.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # skip if we've already processed this event
+        eventDataHash = self.sf.hashstring(event.data)
+        if eventDataHash in self.state["events"]:
+            self.sf.debug(f"Skipping already-processed event, {event.eventType}, from {event.module}")
+            return
+        self.state["events"].append(eventDataHash)
+
+        for hostname in self.query(domain):
+            if hostname.endswith(domain) and not hostname == domain:
```
smicallef, line 139: Please use `SpiderFootTarget.matches()` instead.

modules/sfp_dnsdumpster.py
```diff
@@ -0,0 +1,142 @@
...
+    optdescs = {}
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.sf.debug("Setting up sfp_dnsdumpster")
+        self.state = self.tempStorage()
+        self.state.update({
+            "events": [],
+        })
+        self.__dataSource__ = "DNS"
```
smicallef, line 47: This should not be overridden since technically the data source is DNSDumpster, not DNS directly.

modules/sfp_dnsdumpster.py
```diff
@@ -0,0 +1,142 @@
...
+    opts = {}
+
+    # Option descriptions
+    optdescs = {}
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.sf.debug("Setting up sfp_dnsdumpster")
+        self.state = self.tempStorage()
+        self.state.update({
```
smicallef, line 44: I like the approach taken here but for now can you make this consistent with all other modules (using `self.results`)? We'll move to a better state management approach later on.

modules/sfp_dnsdumpster.py
```diff
@@ -0,0 +1,142 @@
...
+            "events": [],
+        })
+        self.__dataSource__ = "DNS"
+
+        self.opts.update(userOpts)
+
+    def watchedEvents(self):
+        return ["DOMAIN_NAME"]
```
smicallef, line 52: Should probably watch for `INTERNET_NAME` too since subdomains aren't `DOMAIN_NAME`s.

PR #1175, opened 2021-06-03
Feature: Threaded Modules

TheTechromancer:
### This PR improves scan speeds by running each module in its own thread.

threading_vs_nothreading
* **No functionality was deleted, and there are no changes to individual modules.** A non-threaded scan can still be invoked by passing `threaded=False` to `SpiderFootScanner.__startScan()`
threading_arg
* When threading is enabled, each module gets its own dedicated queue for "incoming" events, and shares an "outgoing" queue to `SpiderFootScanner`, which distributes new events across all the modules.
diagram
* A status message periodically shows which modules have the most events queued. This makes it easy to identify slow-performing modules which are bottlenecking the scan.
```
[INFO] sfscan : Events queued: sfp_sslcert: 595, sfp_mnemonic: 569, sfp_pgp: 472, sfp_dnscommonsrv: 400, sfp_fringeproject: 357
```

### DIFF
--- a/sfscan.py
+++ b/sfscan.py
@@ -12,8 +12,12 @@
 import socket
 import sys
 import time
+import queue
+import threading
 import traceback
+from time import sleep
 from copy import deepcopy
+from collections import OrderedDict
 
 import dns.resolver
 
@@ -180,6 +184,9 @@ class SpiderFootScanner():
 
         self.__setStatus("INITIALIZING", time.time() * 1000, None)
 
+        # Used when module threading is enabled
+        self.eventQueue = None
+
         if start:
             self.__startScan()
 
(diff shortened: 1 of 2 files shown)


smicallef:
Awesome contribution - thanks!

bcoles:
Good stuff. Thanks for this.

I notice that scan logs are no longer populated in the `Logs` tab (logs are still shown in terminal). The logging needs to be reworked anyway as part of an ongoing larger piece of work to decouple `sflib` and the `spiderfoot` object. Do you happen to have a quick fix for this in the interim? If not, that's cool, I'll deal with it when I look into reworking the logging.

TheTechromancer:
I'm not sure there's an easy fix for this, since SQLite requires each thread to use its own database handle. I overrode `self.__dbh` for each module which gets around the problem for everything except for logging. Overriding `self.sf._dbh` in a threaded module won't work because every module references the same `SpiderFoot` object.

Here is the code in `spiderfoot/db.py` where it is silently failing:
image

PR #1178, opened 2021-06-12
requirements_test: Add pytest-xdist==2.2.1

bcoles:
I've been using `pytest-xdist` locally for over a year to significantly speed up tests.

I forget if there was a reason we weren't already using this in `requirements_test.txt` for CI. Let's see if the tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - python -m pip install codecov -r requirements_test.txt
 
 script:
-  - python -m pytest --flake8 --durations=5 --cov-report term --cov=. .
+  - python -m pytest --flake8 -n auto --dist loadfile --durations=5 --cov-report term --cov=. .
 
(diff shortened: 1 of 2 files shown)


bcoles:
Ok tests pass. Let's try adding it to Travis.

bcoles:
The tests completed successfully, but only about 16 seconds faster. Compare:

* https://travis-ci.org/github/smicallef/spiderfoot/jobs/774348380
* https://travis-ci.org/github/smicallef/spiderfoot/jobs/774346710

`pytest` correctly identified that two processor cores were available, so I'm not sure why there wasn't much speed up, unless Travis was already performing some multi-core magic. It is also possible that Travis CI infrastructure issues caused discrepancies.

I suspect that over time we'll see a generally lower average completion time.

I see no harm in merging these changes in.

PR #1186, opened 2021-06-14
spiderfoot: Add SpiderFootHelpers

bcoles:
Moves a few functions out of the `SpiderFoot` class in `sflib`. This schizophrenic monster class is an all-encompassing dumping ground for miscellaneous functions and needs to be split up.

The class comment "Common functions used by SpiderFoot modules." has not been accurate for a long time. It is consumed by everything, is expensive to initialise (performs a `deepcopy`, overrides `ssl` properties, overrides the system DNS resolver, ...), and does a bit of everything. This increases merge conflicts, and unnecessarily imports a bunch of dependencies which are rarely needed. More importantly, this also unnecessarily results in tight coupling between components, which in turn complicates development, 

The intention is to move helper functions (ie, anything which doesn't require options set in `sfConfig`) out of `sflib` so we can stop unnecessarily instantiating lots of `SpiderFoot` objects every time we want to call a function, so we can in turn stop passing `sfConfig` around everywhere.

This also helps compartmentalize the code, allowing for cleaner imports and visibility into test coverage, as part of moving towards packaging SpiderFoot.

In the long term it may make sense to further abstract the reporting functionality (JSON, CSV, GEXF) into a separate import as these are rarely needed.

There's much more to be done, but this is a logically stopping point to prevent conflicts with #1175 which is pending merge.

I'm not 100% sold on `SpiderFootHelpers` as a class name. It was easiest cleanest to keep this in a class for now rather than polluting the public namespace. (this is a simple sed recursive find/replace to change later).

Also, the GEXF export functionality is currently broken. This isn't related to the PR - it's due to an update to the `networkx` dependency.


TheTechromancer:
The `@staticmethod` decorator can sometimes be a lifesaver in cases like this, since it allows functions to be called directly from the class (without instantiating it), and doesn't break compatibility.

Anyway just a thought.

PR #1212, opened 2021-06-28
sfwebui: merge stopscan and stopscanmulti

bcoles:


bcoles:
Unit tests and integration tests passed. Only codecov failed during one test.

PR #1239, opened 2021-07-19
Add MassDNS module

TheTechromancer:
This module is the result of my initial rewrite of Spiderfoot's DNSBrute module. After much testing, I found MassDNS to be better-suited for large-scale subdomain brute-forcing.

At its default setting of 1000 concurrent resolvers, this module takes up ~12Mbps bandwidth (up+down) and churns through a 1.2M subdomain wordlist in ~20min.

The advantages of using MassDNS over Python threads are:
- It's faster (easily 1000+ lookups per second, whereas Python's limit is around 200)
- It uses less memory and less CPU
- It doesn't tie up the GIL

This module provides the following features:
- Automatic round-robin across many DNS servers to avoid rate-limiting
- Automatic validation of every nameserver and every newly discovered subdomain
- Automatic handling of wildcard DNS
- Automatic tolerance against network errors, DNS server errors, or garbage data
- Smart alpha mutations such as "-old", "-test", "-prod", as well as sub-variations of existing subdomains
- Smart number mutations which increment/decrement existing numbers in addition to appending 1, 01, 001, etc.
- Numeric, alpha, and subdomain-of-subdomain brute-forcing is applied to `INTERNET_NAME_UNRESOLVED` as well as `INTERNET_NAME`

The subdomain enumeration process goes like this:
1. During module setup, a list of valid servers is retrieved from public-dns.info and only the ones with 99% uptime or higher are selected.
2. Each nameserver is validated by performing one valid query and one garbage query to ensure it responds correctly.
3. For each event, run ShuffleDNS, which uses MassDNS and automatically handles subdomain enumeration and wildcard domains.
4. For each newly-discovered subdomain, validate against known-good DNS servers (`8.8.8.8` and `1.1.1.1`) before triggering an event.

Note: Wildcard entries are filtered out by ShuffleDNS (https://github.com/projectdiscovery/shuffledns).
Note: This module uses a new `ThreadPool` class in `spiderfoot/plugin.py` which integrates with `SpiderFootPlugin`.


TheTechromancer:
Please note that this PR adds two CLI tools (`massdns` and `shuffledns`) to the Dockerfile, which is not updated by the CI/CD tests and therefore causes them to fail. ;)

modules/sfp_massdns.py
```diff
@@ -0,0 +1,405 @@
...
+                nameservers.add(ip_address)
+        # fall back to local dict if necessary
+        if not nameservers:
+            self.sf.debug(f"Failed to retrieve nameservers from {nameservers_url}")
+            nameservers = set(self._ipRegex.findall(open(nameservers_dict, "r").read()))
+            self.sf.debug(f"Loaded {len(nameservers):,} nameservers from {nameservers_dict}")
+        else:
+            self.sf.debug(f"Loaded {len(nameservers):,} nameservers from {nameservers_url}")
+        self.resolvers = self.verifyNameservers(nameservers)
+        self.sf.info(f"Using {len(self.resolvers):,} valid nameservers")
```
bcoles, line 114: The contents of the `resolvers.txt` file is likely to change frequently.

Rather than ship `resolvers.txt` with SpiderFoot, this functionality should be replaced with cacheGet and cachePut.

Dockerfile
```diff
@@ -56,6 +56,16 @@ ENV SPIDERFOOT_LOGS /home/spiderfoot/log
...
+# massdns
+RUN apk --update --no-cache --virtual .build-deps add git build-base go \
+   && git clone --depth=1 https://github.com/blechschmidt/massdns.git \
+   && cd massdns && make && make install
+ENV GOPATH /go
+ENV PATH="$GOPATH/bin:$PATH"
+RUN mkdir -p "$GOPATH/src" "$GOPATH/bin"
+RUN GO111MODULE=on go get -v github.com/projectdiscovery/shuffledns/cmd/shuffledns
+RUN apk del .build-deps
+
```
bcoles, line 68: There is currently no precedence for installing additional third-party tools in the image. However, this is probably a good idea. Perhaps we should look at also installing other tools such as DNStwist. @smicallef will likely have an opinion here.

modules/sfp_massdns.py
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    # Option descriptions
+    optdescs = {
+        "domainonly": "Only brute-force subdomains for the main target (non-recursive).",
+        "numbermutation": "For any host found, increment/decrement existing numbers (if any) and try appending 1, 01, 001, -1, -01, -001, 2, 02, etc. (up to 10)",
+        "alphamutation": "For any host found, try common mutations such as -test, -old, etc.",
+        "large_wordlist": "Use a 1.2M wordlist instead of the default 110K. Takes roughly 20 minutes at 1000 threads.",
+        "concurrent_resolvers": "Maximum concurrent lookup threads. Bandwidth cost is ~1Mbps per 100 resolvers."
+    }
```
bcoles, line 50: Usually modules which leverage third-party tools on the host require the user to specify the file system path in a module option in case the tool is not available in the system environment `PATH`. There's a few reasons to do this, even though this is arguably an environment issue outside of SpiderFoot. @smicallef will likely have an opinion here.

On Windows tools usually aren't installed into the `PATH` by default and often won't add their directory to `PATH`.

There's a security concern due to not using the absolute path. Although only relevant when the system is configured with an insecure `PATH`.

There's a usability concern as there's no error handling in instances where the tool does not exist, making every invocation of the module pointless. Allowing the user to specify a path allows the following error handling:

```python
        if not self.opts['shuffledns_path']:
            self.sf.error("You enabled sfp_tool_massdns but did not set a path to shuffledns!")
            self.errorState = True
            return
```

This will also resolve the failing tests by preventing the test from attempting to execute a file which does note exist:

```
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/subprocess.py:1551: FileNotFoundError
```

modules/sfp_massdns.py
```diff
@@ -0,0 +1,406 @@
...
+        # then, make sure it isn't feeding us garbage data
+        randpool = "bcdfghjklmnpqrstvwxyz3456789"
+        randhost = "".join([random.SystemRandom().choice(randpool) for x in range(10)]) + ".google.com"
+        try:
+            results = list(resolver.query(randhost, "A"))
+            if results:
+                error = f"Nameserver {nameserver} returned garbage data."
+        except Exception:
+            # Garbage query to nameserver failed successfully ;)
+            pass
```
bcoles, line 235: Outside the scope of this PR, but we should probably augment `checkDnsWildcard()` to also take a `resolver` argument rather than duplicate code here.

https://github.com/smicallef/spiderfoot/blob/0a505a836ba7cd6df6a8e3b60605dcbfc757fd41/sflib.py#L2500-L2519

spiderfoot/dicts/README.md
```diff
@@ -0,0 +1,13 @@
+# Spiderfoot Dictionaries / Wordlists
```
bcoles, line 1:
```suggestion
# SpiderFoot Dictionaries / Wordlists
```

modules/sfp_massdns.py
```diff
@@ -0,0 +1,406 @@
...
+        error = None
+
+        resolver = dns.resolver.Resolver()
+        resolver.timeout = timeout
+        resolver.lifetime = timeout
+        resolver.nameservers = [nameserver]
+
+        # first, make sure it can resolve google.com
+        try:
+            resolver.query("www.google.com", "A")
```
bcoles, line 222: This raises warnings:

```
test/unit/modules/test_sfp_massdns.py: 630 warnings
  /home/runner/work/spiderfoot/spiderfoot/modules/sfp_massdns.py:222: DeprecationWarning: please use dns.resolver.Resolver.resolve() instead
    resolver.query("www.google.com", "A")
```

modules/sfp_massdns.py
```diff
@@ -0,0 +1,406 @@
...
+        except Exception:
+            error = f"Nameserver {nameserver} failed to resolve basic query within {timeout} seconds."
+
+        # then, make sure it isn't feeding us garbage data
+        randpool = "bcdfghjklmnpqrstvwxyz3456789"
+        randhost = "".join([random.SystemRandom().choice(randpool) for x in range(10)]) + ".google.com"
+        try:
+            results = list(resolver.query(randhost, "A"))
```
bcoles, line 230: This raises warnings:

```
test/unit/modules/test_sfp_massdns.py: 630 warnings
  /home/runner/work/spiderfoot/spiderfoot/modules/sfp_massdns.py:230: DeprecationWarning: please use dns.resolver.Resolver.resolve() instead
    results = list(resolver.query(randhost, "A"))
```

test/unit/modules/test_sfp_massdns.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_watchedEvents_should_return_list(self):
+        module = sfp_massdns()
+        self.assertIsInstance(module.watchedEvents(), list)
+
+    def test_producedEvents_should_return_list(self):
+        module = sfp_massdns()
+        self.assertIsInstance(module.producedEvents(), list)
+
+    def test_handleEvent(self):
```
bcoles, line 37: This test doesn't test anything and should be skipped.

```suggestion
    @unittest.skip("todo")
    def test_handleEvent(self):
```

Also, once the module supports specifying a path to `shuffledns`, a test can be added to verify that the module fails correctly. For example:

https://github.com/smicallef/spiderfoot/blob/0a505a836ba7cd6df6a8e3b60605dcbfc757fd41/test/unit/modules/test_sfp_tool_cmseek.py#L37-L50

modules/sfp_massdns.py
```diff
@@ -0,0 +1,406 @@
+# -*- coding: utf-8 -*-
```
bcoles, line 1: Modules which leverage third-party tools on the host use a `sfp_tool_` prefix for the module name and file name (`sfp_tool_massdns.py` rather then `sfp_massdns.py`).

bcoles:
> 
> 
> Please note that this PR adds two CLI tools (`massdns` and `shuffledns`) to the Dockerfile. The CICD tests will succeed once the Docker image is updated.

The tests won't pass once the docker file is update as the tests do not use the docker image. The tests fail as the executable is missing. This should be addressed as per https://github.com/smicallef/spiderfoot/pull/1239#discussion_r675950271.

modules/sfp_massdns.py
TheTechromancer: The paths are now autodetected via `shutil.which()` if they are left blank in the settings. Checks were added in `handleEvent` to abort if no executable is present.

smicallef:

Dockerfile
smicallef: I tend to agree with @bcoles here. While I'm keen to have SpiderFoot integrate with more recon tools to expand its reach, bundling the binaries or building of all these tools into the SpiderFoot Docker image feels heavy-handed. It significantly increases the size of the Docker image and makes dependency management more complex. Therefore I'd prefer to keep the building and installation of third party tools separate and up to the user to manage.

smicallef:
I'd prefer that lists coming from other repos are pulled from those repos when SpiderFoot starts up. Otherwise we'll always be maintaining a fork that has to be kept up-to-date.

TheTechromancer:
Both the small and large subdomain wordlists are now dynamically downloaded and cached. The smaller wordlist still exists locally but is only used as a last resort if no others could be loaded.

TheTechromancer:
Dockerfile changes have been reverted.

PR #1244, opened 2021-07-22
In sfp_dnsneighbor, use netmask to respect subnet boundaries

TheTechromancer:
This modifies `sfp_dnsneighbor`  to search the IP's subnet instead of a hardcoded distance in either direction. E.g. with the default netmask of `/28`, the IP of `1.2.3.4` would generate DNS neighbors beginning with `1.2.3.0` and ending with `1.2.3.15`. Before, it would dip down past the subnet boundary into `1.2.2.250`.


smicallef:
Good spot! Thanks!

PR #1246, opened 2021-07-29
Add AbstractAPI Geolocation API

krishnasism:


smicallef:

modules/sfp_abstractapi.py
```diff
@@ -0,0 +1,181 @@
...
+            return None
+
+        if res['code'] == '422':
+            self.sf.error("Usage quota reached. Insufficient API Credit")
+            self.errorState = True
+            return None
+
+        if res['code'] == '429':
+            self.sf.error("Too many requests")
+            self.errorState = True
```
smicallef, line 115: Not sure this should cause `errorState` to be `True` since it indicates one throttled request, not that all future requests will fail (that's the only condition under which `errorState` should be `True`.

krishnasism: Updated

PR #1248, opened 2021-07-30
Nmap + SSL Cert Enhancements

TheTechromancer:
This PR augments the `sfp_tool_nmap` module to generate the following event types in addition to `OPERATING_SYSTEM`:
- `TCP_PORT_OPEN`
- `TCP_PORT_OPEN_SSL` (new type, consumed by `sfp_sslcert`)
- `UDP_PORT_OPEN`
- `LINKED_URL_INTERNAL`

The `sfp_sslcert` module has also been tweaked to consume the new `TCP_PORT_OPEN_SSL` event type, which allows collection of SSL certificates from protocols besides `https`, such as `imaps`, `pop3s`, `ftps`, etc.

`LINKED_URL_INTERNAL` events are now generated whenever a port is found to be using `http` or `https`. This enables `sfp_tool_nmap` to discover websites and services on non-standard ports.

All of this is accomplished with a light service/version detection.
```
# Build Nmap command-line arguments
self.args = (
    # Treat all hosts as online -- skip host discovery
    '-Pn',
    # Never do DNS resolution
    '-n',
    # does the equivalent of --max-rtt-timeout 1250ms --min-rtt-timeout 100ms --initial-rtt-timeout 500ms
    # --max-retries 6 and sets the maximum TCP scan delay to 10 milliseconds.
    '-T4',
    # overrides --max-retries for faster results
    '--max-retries', '2',
    # Probe open ports to determine service/version info
    '-sV',
    # This is a convenience alias for --version-intensity 2.
    # This light mode makes version scanning much faster, but it is slightly less likely to identify services.
    '--version-light'
)
if self.ports:
    self.args += ("-p", ",".join(self.ports))
else:
    self.args += ("--top-ports", str(int(self.opts["topports"])))
# if we're root, enable OS detection
if not (os.name == "posix" and os.geteuid() != 0):
    # Enable OS detection
    self.args += ("-O",)
    # Limit OS detection to promising targets
    self.args += ("--osscan-limit",)
```
Note that this PR adds an Nmap wrapper, `nmappalyzer`, in the form of a Python dependency. This wrapper was written by me, and I chose to make it a separate package because after writing loads of Nmap wrappers, I thought it was finally time to make a library and be done with it. ;)


smicallef:

smicallef:
Nice - I'm generally good with this except it would be good to drop the `Dockerfile` changes from this PR.

PR #1254, opened 2021-08-06
Simplify Dockerfile

TheTechromancer:
This PR simplifies the Dockerfile by using the official `python` image (based off Debian) instead of `alpine`. Alpine's smaller storage footprint is nice but it is not ideal when third-party tools such as Nmap are required.

This PR resolves the following issues I've encountered with the current Docker image:
- Installing packages on an already-built image fails until `/lib/apk/db` is manually recreated
- Installing python packages on an already-built image must be performed as root (the virtual environment is now created under the `spiderfoot` user)
- Tools such as Nmap must be compiled from source, since for reasons unknown, the versions from `apk` do not work.
- Dynamically-linked tools which are dependent on `glibc` (such as the Chrome web driver) will not function on Alpine because it uses `musl`, a lightweight `glibc` alternative.

Python does provide official `slim` and `alpine` images, but they [advise against using them](https://hub.docker.com/_/python) unless image size is the primary concern:
> #### `python:<version>-slim`
> This image does not contain the common packages contained in the default tag and only contains the minimal packages needed to run python. Unless you are working in an environment where only the python image will be deployed and you have space constraints, we highly recommend using the default image of this repository.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@
 # Updated by: Steve Micallef <steve@binarypool.com>
 # Updated by: Steve Bate <svc-spiderfoot@stevebate.net>
 #    -> Inspired by https://github.com/combro2k/dockerfiles/tree/master/alpine-spiderfoot
+# Updated by: TheTechromancer
 #
 # Usage:
 #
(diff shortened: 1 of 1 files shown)


smicallef:
This kind of supports my comment here: https://github.com/smicallef/spiderfoot/pull/1239/files#r684663103

I'd prefer that we don't bundle third party tools and stick with alpine as the base image.

TheTechromancer:
Fair enough. What about an alternate "thicc" Dockerfile that supports all the modules out-of-the-box? I only ask because I am writing other `sfp_tool_*` modules that I think could really benefit from this kind of setup. (I'd be happy to write it of course.)

Docker is such a convenient one-click way to deploy apps; but I think when users realize the level of effort required to get a tool like Nmap running in Alpine, they might either leave the module in a broken state or give up on Docker altogether.

smicallef:
Yeah, I think that's a good approach - a `Dockerfile` for SpiderFoot on its own, and a separate one that pulls down and builds all the tool modules.

PR #1264, opened 2021-08-18
sfwebui: do not show api key instructions for modules with no instructions

bcoles:


bcoles:
Turns out I missed a file when committing.

PR #1272, opened 2021-08-21
Add sfp_abstractapi

bcoles:


bcoles:
I didn't realise there was already an open PR #1246 which implements some of this functionality. Once that module is merged I'll update with the code from this module.

PR #1302, opened 2021-08-28
sfcli: Cleanup

bcoles:
* Minor style cleanup (mostly format string)
* Minor reduction in cyclomatic complexity
* Fix error handling for command line arguments of `int` type (instead of crashing)
* Fix error handling for `shlex` invalid commands (instead of crashing)
* Adds support for `gexf` to `export` command.

### DIFF
--- a/sfcli.py
+++ b/sfcli.py
@@ -355,7 +355,7 @@ class SpiderFootCli(cmd.Cmd):
             return None
 
         if not isinstance(url, str):
-            self.edprint("Invalid request URL: %s" % url)
+            self.edprint(f"Invalid request URL: {url}")
             return None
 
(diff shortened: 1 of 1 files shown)


altjx:
Thanks a lot for this, @bcoles. To get the output in an actual file, I assume we have to use `export <scan_id> -t <export type>` and grab the results from a spooled file?

bcoles:
> Thanks a lot for this, @bcoles. To get the output in an actual file, I assume we have to use `export <scan_id> -t <export type>` and grab the results from a spooled file?

Unfortunately, yes. The export command does not support exporting to file.

I've created PR #1764 to add a `-f [file]` flag for this purpose.

PR #1338, opened 2021-09-04
spiderfoot: event: Remove risk/visibility/confidence args from constructor

bcoles:
We've never used these. They are always set to default.

Other properties that we do actually use, like `actualSource`, are set directly. ie, `event.actualySource = '...'`. We can do the same with risk/visibility/confidence if necessary.


bcoles:
Failed test on MacOS is unrelated.

PR #1342, opened 2021-09-05
Tests: Update acceptance tests

bcoles:
These tests would have caught the HTML rendering issue introduced in #1336 and fixed in #1340.


bcoles:
Failed test is unrelated.

PR #1343, opened 2021-09-05
spiderfoot: store data and cache in user's home directory

bcoles:
This change is important.

It may lead to some initial confusion for anyone using the `master` branch. Short term pain for long term gain.

Your data is fine. Your data in `spiderfoot.db` is fine. It is still there.

The application is loading data from `$HOME/.spiderfoot/spiderfoot.db` now. (or `$SPIDERFOOT_DATA/spiderfoot.db` if you have the environment variable set.)

You can copy/move your old database from `/path/to/spiderfoot/spiderfoot.db` to `$HOME/.spiderfoot/spiderfoot.db` to restore all settings and scans.


smicallef:
Should this include an update to `.gitignore` and `.dockerignore`?

PR #1354, opened 2021-09-14
Add IVRE module 

p-l-:
[IVRE](https://ivre.rocks) is an [open-source](https://github.com/ivre/ivre) network recon framework.

It has several use-cases, such as running your own Shodan-like service (based on powerful open-source tools such as Masscan, Nmap, ZGrab2, ZDns), passively gather intelligence from network traffic (including running a Passive DNS service, collecting and analyzing X509 certificates, HTTP headers, TCP banners, etc.), analyzing scanners hits against simple honeypots, etc.
See for some examples: https://doc.ivre.rocks/en/latest/usage/use-cases.html.


spiderfoot/templates/opts.tmpl
```diff
@@ -140,7 +140,7 @@
                             instructions = ""
                             for step in apiKeyInstructions:
                                 if "https://" in step or "http://" in step:
-                                    step = re.sub(r'.*(https?://.[^ ]+).*', '<a target=_new href=\'\\1\'>\\1</a>', step)
+                                    step = re.sub(r'(https?://.[^ ]+)', '<a target=_new href=\'\\1\'>\\1</a>', step)
```
bcoles, line 143: This change seems reasonable. Submitting this as a separate PR would see the change landed sooner and prevent potential merge conflicts until this PR is reviewed.

As for putting configuration information in `apiKeyInstructions`, usually this would go in the description. Presumably other modules can/will integrate with local services in the future. If it is useful to keep configuration steps separate, it may be worth adding a separate module meta property for this.

p-l-: Thanks for your answer. Submitted as #1355.

I will update this PR to move the instruction to the module description. I also think a specific `dataSource` / `model` (currently set to `"FREE_AUTH_UNLIMITED"`) would be better. What do you think?

p-l-:
@bcoles Thanks for your review. I think it should be better now.

modules/sfp_ivre.py
```diff
@@ -0,0 +1,342 @@
...
+#
+# Created:     2021-09-14
+# Copyright:   (c) Pierre Lalet
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import hashlib
+
+from netaddr import IPNetwork
+from ivre.db import db, DBPassive
```
smicallef, line 16: Any way this could be done using pure REST instead of relying on another library?

modules/sfp_ivre.py
```diff
@@ -0,0 +1,342 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         sfp_ivre
+# Purpose:      Query an IVRE instance
+#
+# Author:      Pierre Lalet <pierre@droids-corp.org>
+#
+# Created:     2021-09-14
+# Copyright:   (c) Pierre Lalet
+# Licence:     GPL
```
smicallef, line 10: Please change to MIT.

modules/sfp_ivre.py
```diff
@@ -0,0 +1,342 @@
...
+
+def flt_from_fqdn(dbase, name):
+    if isinstance(dbase, DBPassive):
+        yield dbase.searchdns(name=name)
+        yield dbase.searchdns(name=name, reverse=True)
+    else:
+        yield dbase.searchhostname(name)
+
+
+EVENTS_FILTERS = {
```
smicallef, line 55: Any particular reason why this approach was taken? It's quite different to all the other modules.

PR #1386, opened 2021-09-21
Docker updates, fixes

TheTechromancer:
Fixes in `Dockerfile.full` and updates to `docker-compose.yml` to accommodate the new data location.

### DIFF
--- a/Dockerfile.full
+++ b/Dockerfile.full
@@ -75,7 +75,7 @@ EXPOSE 5001
 
 # Run the application
 CMD python -c 'from spiderfoot import SpiderFootDb; \
-db = SpiderFootDb({"__database": "spiderfoot.db"}, init=True); \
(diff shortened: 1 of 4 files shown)


Dockerfile.full
```diff
@@ -47,37 +47,36 @@ COPY . .
...
 
 WORKDIR /home/spiderfoot
 
 EXPOSE 5001
 
-# Configure tool paths
-ENTRYPOINT python -c 'from spiderfoot import SpiderFootDb; \
+# Run the application
+CMD python -c 'from spiderfoot import SpiderFootDb; \
 db = SpiderFootDb({"__database": "spiderfoot.db"}, init=True); \
```
bcoles, line 78: The `__database` key is supposed to represent the full file system path.

This should probably be:

```suggestion
db = SpiderFootDb({"__database": "/home/spiderfoot/.spiderfoot/spiderfoot.db"}, init=True); \
```

PR #1389, opened 2021-09-22
SpiderFootDatabase: Add INTERNAL_IP_ADDRESS data type

bcoles:


spiderfoot/db.py
```diff
@@ -161,6 +161,7 @@ class SpiderFootDb:
         ['INTERESTING_FILE', 'Interesting File', 0, 'DESCRIPTOR'],
         ['INTERESTING_FILE_HISTORIC', 'Historic Interesting File', 0, 'DESCRIPTOR'],
         ['JUNK_FILE', 'Junk File', 0, 'DESCRIPTOR'],
+        ['INTERNAL_IP_ADDRESS', 'Internal network IP Address', 0, 'ENTITY'],
```
smicallef, line 164: Should be `IP Address - Internal Network` (all capitalised) so that it's logically ordered in the UI.

PR #1390, opened 2021-09-22
Add Module Shutdown Hook

TheTechromancer:
This PR adds a shutdown hook so that (future) modules that support concurrent/batch processing of events can wrap up their operations cleanly.

### DIFF
--- a/sfscan.py
+++ b/sfscan.py
@@ -423,6 +423,10 @@ class SpiderFootScanner():
             if threaded and not aborted:
                 self.waitForThreads()
 
+            if not threaded:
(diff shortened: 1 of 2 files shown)


sfscan.py
```diff
@@ -466,7 +471,19 @@ def waitForThreads(self):
                         sleep(.1)
                         # but are we really?
                         if self.threadsFinished(log_status):
-                            break
+                            if modulesFinished:
+                                break
+                            else:
```
bcoles, line 476: Redundant `else` after `break`.

TheTechromancer: what would I do without you bcoles

PR #1394, opened 2021-09-23
Censys update

krishnasism:


modules/sfp_censys.py
```diff
@@ -209,8 +189,11 @@ def handleEvent(self, event):
 
             if eventName in ["IP_ADDRESS", "NETBLOCK_OWNER"]:
                 rec = self.queryIp(addr)
-            else:
-                rec = self.queryHost(addr)
+
+            try:
+                rec = rec["result"]
+            except Exception:
+                rec = None
```
bcoles, line 196: This clobbers the existing `rec` variable.

Also `rec` doesn't make sense as a variable name.

Also, you don't need a try/except if you use `.get('result').

You could do:
```suggestion
            if not rec:
                continue

            results = rec.get('results')
            if not results:
                continue
```

This would also require changing every instance of `rec` to `results` in this module.

krishnasism: I didn't work with the variable names, this was a priority item therefore I made sure to just accommodate the newer api functionality to whatever code was present. 

I can do a rec.get('result'), that one's better I guess.
But for the other ones, I put them in the try catch because there's a possibility that the intermediate key might not exist, thus .get will throw a NoneType reference error anyway. So, just made do with try-catch {key error}

krishnasism: Unfortunately, a test failed with rec.get for some reason. I reverted back for now, will take a look at it tomorrow.

PR #1396, opened 2021-09-23
sfp_psbdmp: Remove sfp_psbdmp module

bcoles:
The [API page](https://psbdmp.cc/api) no longer lists the endpoints used by this module.


bcoles:
Looks like the service is still supported. This module may need to be updated for APIv3 instead. Closing.

PR #1488, opened 2021-10-07
SpiderFootHelpers: targetTypeFromString: Add support for bech32 Bitcoin addresses

### LINKED ISSUE #1486: Bitcoin address error
First, I've tried to insert an bitcoin adress containing 40 characteres but I got this error:
Invalid target type. Could not recognize it as a target SpiderFoot supports.

Second, there are some plugins not availables like Bicoin Finder who does web scrapping for a specific address.
How do I add a module to the tool?

bcoles:
Fixes #1486


smicallef:
This would also need a change to `sfp_bitcoin` for completeness.

PR #1535, opened 2021-10-13
Threaded dnsresolve module

TheTechromancer:


Omicron166 (approved):
Legit pull request, should work fine

Omicron166:
@smicallef this pr looks good, pls check it

smicallef:
Similar to my sfp_commonsrv feedback, can we make the number of threads here configurable in the same way other modules do it? A default hard-coded setting of 20 is likely too high for many systems considering how threading-intensive SpiderFoot already is these days.

PR #1536, opened 2021-10-13
Threaded dnscommonsrv module

TheTechromancer:


smicallef:
Sorry it's taken so long to come back to this. Can we make the number of threads here configurable in the same way other modules do it? A default hard-coded setting of 50 is likely too high for many systems considering how threading-intensive SpiderFoot already is these days.

TheTechromancer:
Closing as I have shifted my focus to BBOT.

PR #1581, opened 2021-10-17
removed a pitfall where the code was using a default mutable argument

NaelsonDouglas:
**The problem**
In Python it usually dangerous to use mutable arguments like dicts or lists as default arguments in method, as is better explained [here](https://docs.python-guide.org/writing/gotchas/)

**The solution**
This PR applies a simple refactoring to remove a case where a method was created using default a mutable argument

### DIFF
--- a/sflib.py
+++ b/sflib.py
@@ -2364,7 +2364,7 @@ class SpiderFoot:
 
         return True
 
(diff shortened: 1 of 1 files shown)


bcoles:
This change is reasonable. There are many of these which should be addressed.

smicallef:
Thanks for the contribution!

PR #1582, opened 2021-10-18
Use pretty_uri in results if available

jocejocejoe:
Per WhatsMyName docs:

"pretty_uri" : "if the check_uri is for an API, this OPTIONAL element can show a human-readable page",

so return this instead of the raw API results if is available.


smicallef:
Thanks for the contribution!

PR #1604, opened 2021-11-15
[sfp_greynoise] Suggested Update for GreyNoise Lookup

bradchiappetta:
The following includes some recommended updates to the GreyNoise lookup.  Please feel free to accept or reject as makes sense:

List of changes:
- Migrates individual IP lookups to use IP Lookup endpoints
- Adds support for RIOT dataset - IP lookups check both now
- Updated broken Logo Links
- Update GreyNoise Visualizer URLs
- Corrected display for "GreyNoise" from "Greynoise"

### DIFF
--- a/modules/sfp_greynoise.py
+++ b/modules/sfp_greynoise.py
@@ -203,13 +203,13 @@ class sfp_greynoise(SpiderFootPlugin):
         if "data" in ret and len(ret["data"]) > 0:
             for rec in ret["data"]:
                 if rec.get("seen", None):
-                    self.debug("Found threat info in Greynoise")
+                    self.debug(f"Found threat info in Greynoise: {rec['ip']}")
                     lastseen = rec.get("last_seen", "1970-01-01")
                     lastseen_dt = datetime.strptime(lastseen, "%Y-%m-%d")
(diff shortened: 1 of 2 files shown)


smicallef:
Thanks for this contribution! Is there any way using the Greynoise SDK can be avoided and instead stick with using pure REST? Any additional dependencies can create challenges for various distros to package SpiderFoot, not to mention that we generally want to try and keep the number of dependencies as low as possible.

bradchiappetta:
> Thanks for this contribution! Is there any way using the Greynoise SDK can be avoided and instead stick with using pure REST? Any additional dependencies can create challenges for various distros to package SpiderFoot, not to mention that we generally want to try and keep the number of dependencies as low as possible.

@smicallef : should be all set, I updated everything to use  REST lookups.  Let me know if you need any other changes.

smicallef:
Nice! Thanks!

smicallef:
Looks like a lint check failed - PTAL and I'll merge once tests pass.

bradchiappetta:
> Looks like a lint check failed - PTAL and I'll merge once tests pass.

Okay, I think it is good to go.

PR #1614, opened 2021-12-18
StackOverflow module

JessWill:
Hi! 

I've made a simple StackOverflow module. You can search by domain, and a RAW_RIR_DATA will return for any questions where the domain was seen, with the URL to the question prepended.  Additionally, email and IP addresses are extracted, and the username seen posting the question.

I haven't written python in a long time, but I tried to follow best practices - let me know if I can improve!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python.linting.flake8Enabled": true,
+    "python.linting.enabled": true
+}
--- a/modules/sfp_stackoverflow.py
+++ b/modules/sfp_stackoverflow.py
@@ -69,32 +69,29 @@ class sfp_stackoverflow(SpiderFootPlugin):
     def watchedEvents(self):
         return [
                 "DOMAIN_NAME", 
-                "HUMAN_NAME", 
-                "BITCOIN_ADDRESS", 
-                "COMPANY_NAME",
                 "IP_ADDRESS",
-                "USERNAME"
                 ]
 
     # What events this module produces
     def producedEvents(self):
-        return ["RAW_RIR_DATA", "EMAILADDR"]
+        return ["RAW_RIR_DATA", "EMAILADDR", "USERNAME"]
 
-    def query(self, qry, ):
+    def query(self, qry, qryType):
         # The Stackoverflow excerpts endpoint will search the site for mentions of a keyword and returns a snippet of relevant results
-        res = self.sf.fetchUrl(
-                f"https://api.stackexchange.com/2.3/search/excerpts?order=desc&q={qry}&site=stackoverflow",
+        if qryType == "excerpts":
+            res = self.sf.fetchUrl(
+                    f"https://api.stackexchange.com/2.3/search/excerpts?order=desc&q={qry}&site=stackoverflow",
+                    timeout=self.opts['_fetchtimeout'],
+                    useragent="SpiderFoot"
+                )
+
+        # User profile endpoint
+        if qryType == "questions":
+            res = self.sf.fetchUrl(
+                f"https://api.stackexchange.com/2.3/questions/{qry}?order=desc&sort=activity&site=stackoverflow",
                 timeout=self.opts['_fetchtimeout'],
                 useragent="SpiderFoot"
             )
-
-        # # User profile endpoint
-        # if search == "user":
-        #     res = self.sf.fetchUrl(
-        #         f"https://api.stackexchange.com/2.3/search/excerpts?order=desc&q={qry}&site=stackoverflow",
-        #         timeout=self.opts['_fetchtimeout'],
-        #         useragent="SpiderFoot"
-        #     )
 
         if res['content'] is None:
             self.info(f"No Stackoverflow info found for {qry}")
@@ -106,7 +103,7 @@ class sfp_stackoverflow(SpiderFootPlugin):
             self.error(f"Error processing JSON response from Stackoverflow: {e}")
         return None
 
-    def extractEmail(self, text):
+    def extractEmails(self, text):
         emails = set()
 
         #remove span class highlight, automatically added by stackoverflow to highlight search text
@@ -118,6 +115,20 @@ class sfp_stackoverflow(SpiderFootPlugin):
                 emails.add(match)
 
         return list(emails)
+    
+    def extractUsername(self, questionId):
+        #need to query the questions endpoint with the question_id to find the username
+        query_results = self.query(questionId, "questions")
+        items = query_results.get('items')
+
+        if query_results is None:
+            return 
+
+        for item in items:
+            owner = item['owner']
+            username = owner.get('display_name')
+        
+        return str(username)
 
     def handleEvent(self, event):
         eventName = event.eventType
@@ -132,16 +143,16 @@ class sfp_stackoverflow(SpiderFootPlugin):
 
         self.results[eventData] = True
 
-        query_results = self.query(eventData)
+        query_results = self.query(eventData, "excerpts")
         items = query_results.get('items')
-
         allEmails = []
+        allUsernames = []
 
         #iterate through all results from query, creating raw_rir_data events and extracting emails      
         for item in items:
             if self.checkForStop():
                 return
-
+            
             # return raw_rir_data event
             body = item["body"]
             excerpt = item["excerpt"]
@@ -150,17 +161,25 @@ class sfp_stackoverflow(SpiderFootPlugin):
                                 str("<SFURL>https://stackoverflow.com/questions/")+str(question)+str("</SFURL>")+str("\n")+str(body)+str(excerpt), 
                                 self.__name__, event)
             self.notifyListeners(e)
-
-            #Extract email addresses and add to the allEmails list
+            
             text = body+excerpt
-            emails = self.extractEmail(text)
+            #Extract other interesting events
+            emails = self.extractEmails(text)
             allEmails.append(emails)
 
-        if emails:
-            for email in emails:
+            questionId = item["question_id"]
+            username = self.extractUsername(questionId)
+            allUsernames.append(username)
+
+        if allEmails:
+            for email in allEmails:
+                email = str(email)
                 e = SpiderFootEvent('EMAILADDR', email, self.__name__, event)
                 self.notifyListeners(e)
 
+        if allUsernames:
+            for username in allUsernames:
+                e = SpiderFootEvent('USERNAME', username, self.__name__, event)
+                self.notifyListeners(e)
 
-        
 # End of sfp_stackoverflow class
--- a/modules/sfp_stackoverflow.py
+++ b/modules/sfp_stackoverflow.py
@@ -20,9 +20,9 @@ from spiderfoot import SpiderFootEvent, SpiderFootPlugin
 class sfp_stackoverflow(SpiderFootPlugin):
 
     meta = {
-        'name': "Stackoverflow Module",
-        'summary': "Search StackOverflow for any mentions of your target. Returns potentially related information.",
-        'flags': ["apikey","errorprone"],
+        'name': "Stackoverflow",
+        'summary': "Search StackOverflow for any mentions of a target domain. Returns potentially related information.",
+        'flags': ["errorprone"],
         'useCases': ["Passive"],
         'categories': ["Content Analysis"],
         'dataSource': {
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
smicallef, line 1: This file shouldn't be included.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         sfp_stackoverflow
+# Purpose:      Example module to use for new modules.
```
smicallef, line 4: Please update this to be descriptive of this module.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+            'model': "FREE_NOAUTH_LIMITED",
+            'references': [
+                "https://api.stackexchange.com/docs"
+            ],
+            'apiKeyInstructions': [
+                "Visit https://stackapps.com/apps/oauth/register",
+            ],
+            'favIcon': "https://cdn.sstatic.net/Sites/stackoverflow/Img/favicon.ico?v=ec617d715196",
+            'logo': "https://cdn.sstatic.net/Sites/stackoverflow/Img/apple-touch-icon.png",
+            'description': "A paragraph of text with details about the data source / services. "
```
smicallef, line 40: This should not be the boilerplate text.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    # Default Options
+    opts = {
+        'api_key': '',
+    }
+
+    # Option descriptions
+    optdescs = {
+        "api_key": "StackApps Optional API Key."
```
smicallef, line 54: What limits are imposed if not provided? Would be good to have a few words about that here (nit: Optional should be all lower case).

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    # Results Tracking
+    results = None
+
+    # Tracking the error state of the module
+    errorState = False
```
smicallef, line 61: errorState doesn't appear to be used in the module. This is typically used in cases where the module encounters an error that should mean it no longer processes any events, e.g. if the API appears to be down or credentials rejected.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+        except Exception as e:
+            self.error(f"Error processing JSON response from Stackoverflow: {e}")
+        return None
+
+    def extractEmails(self, text):
+        emails = set()
+
+        # Remove span class highlight, automatically added by stackoverflow to highlight search text
+        newText = text.replace("<span class=\"highlight\">", "")
+        matches = re.findall(r'([\%a-zA-Z\.0-9_\-\+]+@[a-zA-Z\.0-9\-]+\.[a-zA-Z\.0-9\-]+)', newText)
```
smicallef, line 110: Better to use the `parseEmails()` function in `sflib`.

JessWill:
Thanks @smicallef! I'll get to fixing that all up today.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+        # Iterate through all results from query, creating raw_rir_data events and extracting emails
+        for item in items:
+            if self.checkForStop():
+                return
+
+            # create raw_rir_data event
+            body = item["body"]
+            excerpt = item["excerpt"]
+            question = item["question_id"]
+            e = SpiderFootEvent('RAW_RIR_DATA',
```
smicallef, line 174: `RAW_RIR_DATA` is typically for the raw API response.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+        return str(username)
+
+    def extractIPs(self, text):
+        ips = set()
+
+        matches = re.findall(r'^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$', text)
+
+        if matches:
+            for match in matches:
+                if self.sf.validIP(match) and not(netaddr.IPAddress(match).is_loopback()):
```
smicallef, line 141: You can use `isValidLocalOrLoopbackIp()` in sflib here.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,208 @@
...
+        for item in items:
+            owner = item['owner']
+            username = owner.get('display_name')
+
+        return str(username)
+
+    def extractIPs(self, text):
+        ips = set()
+
+        matches = re.findall(r'^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$', text)
```
smicallef, line 137: Is there a chance of IPv6 addresses by any chance?

modules/sfp_stackoverflow.py
JessWill: Would you like me to return the raw API response instead? It's just not very aesthetically pleasing in the UI and contains a lot of unnecessary info. Is there an alternative Event for interesting text? Could I even just return a URL to the question where the domain was seen?

smicallef: Yeah the RAW API response is preferred, reason being that some modules look at the event to extract hostnames and potentially other data of interest.

modules/sfp_stackoverflow.py
JessWill: Done! :)

smicallef:
Thanks for all the improvements! Just a couple of linter fixes and should be good to merge.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,224 @@
...
+        'useCases': ["Passive"],
+        'categories': ["Content Analysis"],
+        'dataSource': {
+            'website': "https://www.stackecxchange.com",
+            'model': "FREE_NOAUTH_LIMITED",
+            'references': [
+                "https://api.stackexchange.com/docs"
+            ],
+            'apiKeyInstructions': [
+                "Visit https://stackapps.com/apps/oauth/register",
```
smicallef, line 35: Could we have a few more steps here to explain the process? I'm presented with an OAuth screen here and struggling to understand what gets me to the API key.

smicallef:
I tested the module against `tesla.com` without an API key, and there are a couple of issues:
1) email addresses returned are in the format of `['elon@spaceX.com']` for example.
2) Possibly due to the above, but email addresses on domains other than those of the target should be `AFFILIATE_EMAILADDR`.
3) Values for `USERNAME` are not always usernames, some are full names or initials, etc. so this needs to be cleaned up.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,255 @@
...
+            email = str(email)
+            if eventData in email:
+                e = SpiderFootEvent('EMAILADDR', email, self.__name__, event)
+                self.notifyListeners(e)
+            else:
+                e = SpiderFootEvent('AFFILIATE_EMAILADDR', email, self.__name__, event)
+                self.notifyListeners(e)
+
+        for username in allUsernames:
+            e = SpiderFootEvent('USERNAME', username, self.__name__, event)
```
smicallef, line 243: I'm still seeing non-username types for usernames. If you test against `tesla.com`, you'll see usernames like `A J E` and `stacj aa`. Probably anything with a space should be disregarded, but I'm also not sure how good the quality of usernames are here.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,255 @@
...
+                allIP4s.append(ip4s)
+
+            ip6s = self.extractIP6s(text)
+            if ip6s is not None:
+                allIP6s.append(ip6s)
+
+        # create events for emails, username and IPs
+        for email in allEmails:
+            email = str(email)
+            if eventData in email:
```
smicallef, line 235: This could result in a lot of false positives. Use `self.getTarget().matches()` instead.

modules/sfp_stackoverflow.py
JessWill: All of that information is pulled directly from the displayname value in the API response:
image

Users on StackOverflow are allowed to set a username with spaces, "stacj aa" is somebodies actual username
https://stackoverflow.com/users/14085501/stacj-aa?tab=topactivity

If I remove anything with a space, we'll miss out on legitimate users

modules/sfp_stackoverflow.py
JessWill: The email will never match the target exactly, so I've opted to use:

   if email.endsWith(self.getTarget()):

 Does that seem okay?

smicallef: Hmm that won't work, since getTarget() may also be an IP address or something else entirely. If you pass the domain of the email to `matches()`, it should be smart enough to handle this correctly.

modules/sfp_stackoverflow.py
smicallef: SpiderFoot internally as well as the modules processing `USERNAME` events may encounter issues if spaces are contained. Maybe if it has a space, report it instead as a "Potential full name: <username>" `RAW_RIR_DATA` event for `sfp_names` to handle?

JessWill: Cool! I think that's should be sorted now - looks like the human names are all correctly being reported and all of the usernames don't have spaces and also aren't human names. 👍

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,257 @@
...
+            self.opts[opt] = userOpts[opt]
+
+    # What events is this module interested in for input
+    def watchedEvents(self):
+        return ["DOMAIN_NAME"]
+
+    # What events this module produces
+    def producedEvents(self):
+        return ["RAW_RIR_DATA",
+                "EMAILADDR"
```
smicallef, line 79: Missing comma at the end here.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,257 @@
...
+from modules.sfp_names import sfp_names
+
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_stackoverflow(SpiderFootPlugin):
+
+    meta = {
+        'name': "Stackoverflow",
```
smicallef, line 24: Probably should be StackOverflow?

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,257 @@
...
+
+            ip6s = self.extractIP6s(text)
+            if ip6s is not None:
+                allIP6s.append(ip6s)
+
+        # create events for emails, username and IPs
+        for email in allEmails:
+            email = str(email)
+            if self.getTarget().matches(email):
+                e = SpiderFootEvent('EMAILADDR', email, self.__name__, event)
```
smicallef, line 224: I see mixed-case stuff appearing, so would be better to `.lower()` the `email`.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,257 @@
...
+        return ["RAW_RIR_DATA",
+                "EMAILADDR",
+                "AFFILIATE_EMAILADDR",
+                "USERNAME",
+                "IP_ADDRESS",
+                "IPV6_ADDRESS",
+                "HUMAN_NAME"
+                ]
+
+    def query(self, qry, qryType):
```
smicallef, line 87: In my testing I'm seeing throttling errors returned, indicated by a return code of `502`. This should be handled, with an error log reported. I don't know what the throttling limits are depending on whether an API key is provided, but perhaps have a `sleep` after each call to avoid throlling.

modules/sfp_stackoverflow.py
```diff
@@ -0,0 +1,257 @@
...
+
+        if matches:
+            for match in matches:
+                if self.sf.validIP6(match) and not(self.sf.isValidLocalOrLoopbackIP(match)):
+                    ips.add(match)
+            return list(ips)
+        else:
+            return None
+
+    def handleEvent(self, event):
```
smicallef, line 166: For all events reported, there should be some checking to avoid reporting the same event twice. For example, testing with `tesla.com` I'm seeing the same email address being reported multiple times.

smicallef:
Just a few more points of feedback. Sorry for dragging this on, but it's looking great and should be ready to merge once these points are addressed.

JessWill:
> 

No problems at all, it's good experience for me!  Happy to make changes

Issue #1612, opened 2021-12-11
Unexpected behavior of dark mode toggle on Chromium


Rb64:
When using a Chromium based browser the dark mode toggle doesn’t beave as expected. The Toggle is not set in Dark mode. This problem is not existent in Firefox. I also noticed that the text next to the toggle is set to “Light Mode” when the theme is dark and is set to “Dark Mode” when it is Light, is this intentional? I have observed this second potential issue on Firefox as well.
![grafik](https://user-images.githubusercontent.com/91498309/145683157-39b050f0-ecc1-45d2-bcdf-6e5e9ba8e453.png)

PR #1615, opened 2021-12-19
Darkmodefix Issue #1612

Rb64:
Fixed problems mentioned in Issue #1612. I fixed the dark mode toggle on chromium-based browsers and ensuring that the toggle responds as expected. I also corrected the Label next to the toggle so that it displays "Dark Mode" when the theme is in Dark Mode and "Light Mode" otherwise.


smicallef:
Thanks for the fix!

PR #1617, opened 2021-12-27
Added ability to change the number of concurrent threads

hanzala123:
While Using Spiderfoot on a project i needed to crank up it's speed. High resource usage wasn't an issue. But i couldn't find a way to to that using the switches. So i added an option here to do that.


smicallef:
Nice improvement, thanks! Just a small flake to fix and then I'll merge.

PR #1618, opened 2021-12-29
Allow username and password to be set as environment variables

cs4p:
This makes it a bit easier to run a secure instance in a Kubernetes cluster or other virtualized environments.  I also had to change the requirements file to load a specific version of networkx and secure packages.  Seems like the newest versions are not compatible.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,11 @@ services:
   spiderfoot:
     build:
        context: ./
-    image: spiderfoot
+    image: my_spiderfoot
     container_name: spiderfoot
     ports:
       - "5001:5001"
+    environment:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -8,4 +8,7 @@ services:
     container_name: spiderfoot
     ports:
       - "5001:5001"
+    environment:
+      - SF_USERNAME=admin
+      - SF_PASSWORD=admin
```
smicallef, line 13: I would rather not encourage having passwords in environment variables.

smicallef:
Thanks for the contribution, but I'd rather not support setting credentials as environment variables.

PR #1645, opened 2022-04-12
Create outlier_web_server_header.yaml

geeknik:
Outlier detection -- A web server produced a header response which appeared in 10% or less of the total web servers found.

### DIFF
--- a/correlations/outlier_web_server_header.yaml
+++ b/correlations/outlier_web_server_header.yaml
@@ -1,0 +1,19 @@
+id: outlier_web_server_header
+version: 1
+meta:
+  name: Outlier web server header
+  description: >
+    A web server produced a header response which appeared
+    in 10% or less of the total web servers found.
+  risk: INFO
+collections:
+  - collect:
+      - method: exact
+        field: type
+        value: WEBSERVER_HTTPHEADERS
+aggregation:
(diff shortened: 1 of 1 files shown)


smicallef:
Thanks for the submission! One question: how does this provide value over the `outlier_webserver` rule? I think this might be prone to be quite noisy since it's looking at the `WEBSERVER_HTTPHEADERS` type which can be very unique (e.g. due to timestamps, hashes, etc.) without actually telling you much about the target.

geeknik:
Well I was hopeful that by using `WEBSERVER_HTTPHEADERS` it would only look at the header name, not the value. I'm more concerned about the name of the header vs the value of the header. But since I've never seen my correlation rule fire despite using it on quite a few scans with 150K or more elements, I wasn't sure what kind of data was going to be returned.

smicallef:
I suspect this rule hasn't fired anything for you locally because the headers are too anomalous (because values are considered, not just the header, and even if we only looked at headers the headers are reported in aggregate, not individually), and the correlation engine has a safety mechanism for reporting outliers when the dataset is highly anomalous. I'd therefore suggest closing this PR until we implement some functionality to the engine that manipulates the data for processing.

PR #1651, opened 2022-04-26
Fixed smallest typo for stackecxchange

mandreko-ts:
I was tinkering with spiderfoot and ran across this typo which was the smallest of all annoyances anyone could ever experience. However, I figured it should be fixed.


smicallef:
Thanks!

PR #1657, opened 2022-05-01
Modules: Keep module meta useCases array items on single line

bcoles:
This makes them grepable.

Helping users on GitHub and Discord has required grepping modules on dozens of occasions.


Omicron166 (approved):
Why not?

smicallef:
> Why not?

Helpful when grepping through module files to find which are in a given use case.

PR #1659, opened 2022-05-01
sfp_stackoverflow: Cleanup and add tests

bcoles:
This fixes a design flaw in the `sfp_stackoverflow` module. The module tries to setup the `sfp_names` module and force it to handle an event. This goes against the SpiderFoot architecture. This is also unnecessary if the `sfp_names` module was included in the scan. If this was any other module it would usually be a minor inconvenience. Unfortunately, the `sfp_names` module loads multiple dictionaries during startup, resulting in a `list` of almost 900 thousand entries every time the module encountered a username. This causes issues on systems with ~2GB RAM or less.

Also reports IP addresses as affiliates `AFFILIATE_IPADDR` / `AFFILIATE_IPV6_ADDRESS` rathar then `IP_ADDRESS` / `IPV6_ADDRESS`.

### DIFF
--- a/modules/sfp_stackoverflow.py
+++ b/modules/sfp_stackoverflow.py
@@ -13,8 +13,6 @@
 import json
 import re
 import time
-from modules.sfp_names import sfp_names
-
 
 from spiderfoot import SpiderFootEvent, SpiderFootPlugin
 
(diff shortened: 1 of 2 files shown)


bcoles:
Tests failed due to DNS as usual. Unfortunately none of the other tests ran before they were cancelled.

PR #1664, opened 2022-05-03
Threat Jammer module for blacklist/malicious IP

diegoparrilla:
# ThreatJammer module
<p align="center">

</p>

### Why a module for Threat Jammer?

[ThreatJammer.com](https://threatjammer.com/) is a service that allows developers, security engineers, and other IT professionals to access high-quality threat intelligence data from a variety of sources and integrate it into their applications with the sole purpose of detecting and blocking malicious activity.

The assessment endpoint of the API returns a risk score with the estimated probability of being a malicious IP address. Less than 35 means Low risk, and more than 67 means High risk. This is the API endpoint used in the module for IPv4 and IPv6 addresses for the BLACKLISTED_IPADDR and MALICIOUS_IPADDR events.

Anybody can sign up and use the service for free forever. 

### Quick start

To use the module the user must [sign up and obtain an API KEY](https://threatjammer.com/docs/threat-jammer-api-keys). The sign-up is free forever. 

Once the developers have obtained an API Key, they have to enable the module in the `Settings` section and populate the options:
- *API hostname*: [Mandatory] The hostname where the API endpoint runs. By default `dublin.api.threatjammer.com`. 
- *API key*: [Mandatory] The API key obtained when signing up.
- *Minimum Threat Jammer risk score*: [Mandatory] The minimum integer from 0 to 99 is the threshold of the risk sensitivity. An IP address with less score than the given will be discarded and not displayed. 

Don't forget to click on the `Save Changes` button.

### Python and OS versions

The code has passed the tests implemented in the CI workflows, as expected. This module is compatible with versions of Python from v3.7 and up.

This module should be considered a beta version.

### DIFF
--- a/modules/sfp_threatjammer.py
+++ b/modules/sfp_threatjammer.py
@@ -1,0 +1,205 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_threatjammer
+# Purpose:     Check if an IP or netblock is malicious according to ThreatJammer.com.
+#
+# Author:      diego.parrilla.santamaria@gmail.com
+#
+# Created:     05/05/2022
+# Copyright:   (c) Diego Parrilla 2022
+# Licence:     GPL
+# -------------------------------------------------------------------------------
+
+import json
+import time
+
+from spiderfoot import SpiderFootEvent, SpiderFootPlugin
+
+
+class sfp_threatjammer(SpiderFootPlugin):
+
+    meta = {
+        'name': "Threat Jammer",
+        'summary': "Check if an IP or netblock is malicious according to ThreatJammer.com",
+        'flags': ["apikey"],
+        'useCases': ["Passive", "Investigate"],
+        'categories': ["Reputation Systems"],
+        'dataSource': {
+            'website': "https://threatjammer.com",
+            'model': "FREE_AUTH_LIMITED",
+            'references': [
+                "https://threatjammer.com/docs/what-is-threat-jammer",
+                "https://threatjammer.com/docs/how-threat-jammer-works",
+                "https://threatjammer.com/docs/introduction-threat-jammer-user-api",
+                "https://threatjammer.com/docs/introduction-threat-jammer-report-api",
+                "https://threatjammer.com/tutorials/how-to-configure-fail2ban-in-ubuntu",
+                "https://threatjammer.com/tutorials/how-to-configure-cowrie-honeypot"
+            ],
+            'apiKeyInstructions': [
+                "https://threatjammer.com/docs/threat-jammer-api-keys",
+                "Register a new account with an email",
+                "Navigate to https://threatjammer.com/keys",
+                "The API Keys are listed under 'API Key' column"
+            ],
+            'favIcon': "https://threatjammer.com/favicon.ico",
+            'logo': "https://threatjammer.com/threatjammer-risk-score.png",
+            'description': "Threat Jammer is a service to access high-quality threat intelligence"
+            " data from a variety of sources, and integrate it into their applications with the"
+            " sole purpose of detecting and blocking malicious activity."
+        }
+    }
+
+    opts = {
+        'api_key': '',
+        "risk_score_min": 35,
+        'checkaffiliates': True,
+    }
+
+    optdescs = {
+        'api_key': "ThreatJammer.com API key.",
+        "risk_score_min": "Minimum Threat Jammer risk score.",
+        'checkaffiliates': "Apply checks to affiliates?",
+    }
+
+    results = None
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.results = self.tempStorage()
+
+        for opt in list(userOpts.keys()):
+            self.opts[opt] = userOpts[opt]
+
+    def watchedEvents(self):
+        return [
+            "IP_ADDRESS",
+            "IPV6_ADDRESS",
+            "AFFILIATE_IPADDR",
+            "AFFILIATE_IPV6_ADDRESS",
+        ]
+
+    def producedEvents(self):
+        return [
+            "BLACKLISTED_IPADDR",
+            "BLACKLISTED_AFFILIATE_IPADDR",
+            "MALICIOUS_IPADDR",
+            "MALICIOUS_AFFILIATE_IPADDR",
+        ]
+
+    def queryIp(self, ip):
+        """Query API for an IPv4 or IPv6 address.
+
+        Args:
+            ip (str): IP address
+
+        Returns:
+            str: API response as JSON
+        """
+
+        headers = {
+            'Authorization': f"Bearer {self.opts['api_key']}",
+            'Accept': 'application/json',
+        }
+
+        res = self.sf.fetchUrl(
+            f"https://dublin.api.threatjammer.com/v1/assess/ip/{ip}",
+            timeout=self.opts['_fetchtimeout'],
+            useragent=self.opts['_useragent'],
+            headers=headers
+        )
+
+        time.sleep(1)
+
+        if res['code'] == '429':
+            self.error("You are being rate-limited by ThreatJammer.com")
+            self.errorState = True
+            return None
+
+        if res['code'] != "200":
+            self.error("Error retrieving search results from ThreatJammer.com")
+            self.errorState = True
+            return None
+
+        if res['content'] is None:
+            self.error("Received no content from ThreatJammer.com")
+            self.errorState = True
+            return None
+
+        try:
+            return json.loads(res['content'])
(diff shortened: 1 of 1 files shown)


modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_threatjammer
+# Purpose:     Check if an IP or netblock is malicious according to ThreatJammer.com.
+#
+# Author:      diego.parrilla.santamaria@gmail.com
+#
+# Created:     05/05/2022
```
bcoles, line 8: International date format is preferred, future man.

```suggestion
# Created:     2022-05-03
```

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
...
+        'checkaffiliates': True,
+    }
+
+    optdescs = {
+        'api_key': "ThreatJammer.com API key.",
+        "risk_score_min": "Minimum Threat Jammer risk score.",
+        'checkaffiliates': "Apply checks to affiliates?",
+    }
+
+    results = None
```
bcoles, line 64:
```suggestion
    results = None
    errorState = False
```

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.results = self.tempStorage()
```
bcoles, line 68:
```suggestion
        self.results = self.tempStorage()
        errorState = False
```

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+        time.sleep(1)
+
+        if res['code'] == '429':
+            self.error("You are being rate-limited by ThreatJammer.com")
+            self.errorState = True
+            return None
+
+        if res['code'] != "200":
```
bcoles, line 118: Does the server reply with an error when the provided API key is invalid (401 / 403) or quota ? If so, this function should handle the appropriate code explicitly and set `errorState` with an appropriate error, rather than presume all non-200 codes are fatal.

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
...
+        if res['content'] is None:
+            self.error("Received no content from ThreatJammer.com")
+            self.errorState = True
+            return None
+
+        try:
+            return json.loads(res['content'])
+        except Exception as e:
+            self.debug(f"Error processing JSON response: {e}")
+            return None
```
bcoles, line 132:
```suggestion
```

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,205 @@
...
+        self.debug(f"Checking maliciousness of IP address {eventData} with ThreatJammer.com")
+
+        ip_info = self.queryIp(eventData)
+
+        if ip_info is None:
+            self.sf.error(f"Error processing JSON response for {eventData} from ThreatJammer.com")
+            return
+
+        risk_score = int(ip_info["score"])
+        risk = ip_info["risk"]
```
bcoles, line 178: These should be checked for validity.

Something like this (depending on how zero is handled).

```suggestion
        risk_score = ip_info.get('score')
        if not risk_score:
            halt and catch fire
            return
        risk_score = int(risk_score)

        risk = ip_info('risk')
        if not risk:
            everything is okay. maybe print a debug message.
            return
```

bcoles (commented):
I gave the module a once over and it seems reasonable. A few minor observations.

modules/sfp_threatjammer.py
diegoparrilla: Time-machine fixed

modules/sfp_threatjammer.py
diegoparrilla: Added better error handling

modules/sfp_threatjammer.py
diegoparrilla: Done

modules/sfp_threatjammer.py
diegoparrilla: Implemented check for broken JSON returned (no score or risk)

diegoparrilla:
The new version implements the changes suggested. It also implements a new optional parameter with the API hostname.

bcoles:
@smicallef please start the CI workflow.

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,238 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:        sfp_threatjammer
+# Purpose:     Check if an IP or netblock is malicious according to ThreatJammer.com.
+#
+# Author:      diego.parrilla.santamaria@gmail.com
+#
+# Created:     2022-05-03
+# Copyright:   (c) Diego Parrilla 2022
```
smicallef, line 9: Please change the copyright to `Steve Micallef`. This is required as I need to retain copyright control in the event that I change from GPL to something else.

smicallef, line 9: Please change the copyright to `Steve Micallef`. This is required as I need to retain copyright control in the event that I change from GPL to something else.

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def setup(self, sfc, userOpts=dict()):
+        self.sf = sfc
+        self.results = self.tempStorage()
+        self.errorState = False
+
+        for opt in list(userOpts.keys()):
+            self.opts[opt] = userOpts[opt]
+
+    def watchedEvents(self):
```
smicallef, line 77: Should this module also receive various netblock events and process those as various other similar modules do?

smicallef, line 77: Should this module also receive various netblock events and process those as various other similar modules do?

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,238 @@
...
+            return None
+
+        if res['code'] == '400':
+            self.error("ThreatJammer.com rejected the IP address. Use only public IP addresses.")
+            self.errorState = True
+            return None
+
+        if res['code'] == '422':
+            self.error("ThreatJammer.com could not process the IP address. Check the format.")
+            self.errorState = True
```
smicallef, line 134: Note that `errorState` should only be used in cases where the module should no longer be invoked for other events. It's used in cases where the upstream or module should no longer be called due to some kind of critical failure.

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,238 @@
...
+        'api_hostname': "dublin.api.threatjammer.com",
+        'risk_score_min': 35,
+        'checkaffiliates': True,
+    }
+
+    optdescs = {
+        'api_key': "API key",
+        'api_hostname': "User API hostname",
+        'risk_score_min': "Minimum Threat Jammer risk score",
+        'checkaffiliates': "Apply checks to affiliates?",
```
smicallef, line 63: Did you want to have `checkcohost` functionality like other similar modules have?

smicallef, line 63: Did you want to have `checkcohost` functionality like other similar modules have?

modules/sfp_threatjammer.py
```diff
@@ -0,0 +1,238 @@
...
+            self.sf.error(f"Error processing JSON response for {eventData} from ThreatJammer.com")
+            return
+
+        score = ip_info.get('score')
+        if not score:
+            self.sf.error(f"No risk score found for {eventData} from ThreatJammer.com. Skipping.")
+            return
+        risk_score = int(score)
+
+        risk = ip_info["risk"]
```
smicallef, line 208: Perhaps `.get()` instead?

smicallef, line 208: Perhaps `.get()` instead?

diegoparrilla:
Good feedback! I have to review some of the suggestions because I don't understand the implications, I will try to do my best this weekend 👍

modules/sfp_threatjammer.py
diegoparrilla: Not in this version of the Threat Jammer API

modules/sfp_threatjammer.py
diegoparrilla: I don't think so. The current version of the API only performs IP reputation/risk scoring.

diegoparrilla (commented):
Suggested changes applied.

modules/sfp_threatjammer.py
smicallef: Any reason this was resolved? Most other modules in this category have such functionality and it would be good to have it from the beginning.

PR #1665, opened 2022-05-04
Tests: Unit: Modules: Add tests for parseApiResponse function

bcoles:


bcoles:
All tests passed except Python 3.7 on Ubuntu which appeared to hang.

I noticed a couple of the other recent PRs have also hung similarly. The tests pass on most systems, then one gets stuck at 97% and eventually times out.

This behaviour isn't related to this PR. The first occurance of a hang at 97% happened in 000a75a57f856e5e4109708c38a70172e21d399f here:

* https://github.com/smicallef/spiderfoot/runs/5862820602?check_suite_focus=true

But the issue may have started occurring earlier. It is hard to tell where the issue was first introduced. It does not appear to be related to a specific combination of Python and OS.

The tests pass locally on Python 3.8 and 3.9.

PR #1666, opened 2022-05-04
Tests: Do not fail CI tests if codecov fails

bcoles:


bcoles:
Tests failed due to module DNS tests again.

PR #1673, opened 2022-05-07
sflib: Code cleanup

bcoles:


bcoles:
The `dictnames` and `dictwords` functions return a `set()` and are only ever used as a `set()`. Each call to `dictwords` created a temporary `dict()` object and returned `set(dict.keys())`.

Temporarily storing these words as `dict()` keys was used for de-duplication. This is unnecessary and can be handled natively using `set()`. Removing the use of `dict()` should reduce memory usage slightly and saves a call to `set()` at the end. Memory management is more important now that SpiderFoot is multi-threaded, as several modules (`sfp_accounts`, `sfp_names`, and `sfp_binstring`) could run concurrently, resulting in multiple concurrent calls to `dictwords()`.

Memory usage while running unit and integration tests (excluding module integration tests) on a single core shown below. Although this probably isn't the best way to measure memory usage due to Python's garbage collection.

The change to memory usage is minimal, but the spikes look cleaner. Also, note that the Y axis scale is not always identical in the graphs. This is because the usage sometimes exceeded 700MB.

## Before

before

before

## After

after

after

PR #1684, opened 2022-05-23
Corrected installation of Google Chrome

anoduck:
Apt-key is deprecated and failed to install the key needed for installation of Google Chrome. Removed apt-key command and added the new way to fetch keys in Debian, which is a good ole-fashioned wget and gpg --dearmour.


smicallef:
Thanks for the contribution!

anoduck:
No, Thank you, for the incredible-ness that is spiderfoot. This is my directly accepted pull request, a meaningful achievement for my humble beginnings. I am so totally stoked about it.

PR #1687, opened 2022-05-28
sflib: Remove myPath() and move wordlist functions to SpiderFootHelpers

bcoles:


sf.py
```diff
@@ -61,7 +61,7 @@ def main() -> None:
         '_fetchtimeout': 5,  # number of seconds before giving up on a fetch
         '_internettlds': 'https://publicsuffix.org/list/effective_tld_names.dat',
         '_internettlds_cache': 72,
-        '_genericusers': "abuse,admin,billing,compliance,devnull,dns,ftp,hostmaster,inoc,ispfeedback,ispsupport,list-request,list,maildaemon,marketing,noc,no-reply,noreply,null,peering,peering-notify,peering-request,phish,phishing,postmaster,privacy,registrar,registry,root,routing-registry,rr,sales,security,spam,support,sysadmin,tech,undisclosed-recipients,unsubscribe,usenet,uucp,webmaster,www",
+        '_genericusers': ",".join(SpiderFootHelpers.usernamesFromWordlists(['generic-usernames'])),
```
smicallef, line 64: Is the user overrides this in the UI, they will get lost the next time SpiderFoot is restarted, right?

bcoles: This behaviour is identical to the existing behaviour.

smicallef: My mistake, thanks.

PR #1750, opened 2023-01-29
Update target.py

MiChaelinzo:
Changed variable names to follow snake_case naming convention Removed unnecessary comments
Simplified the validation for targetType and targetValue by using the if not statement Merged the getAddresses function and added the validation for IPV6_ADDRESS Merged the getNames function

### DIFF
--- a/spiderfoot/target.py
+++ b/spiderfoot/target.py
@@ -1,214 +1,57 @@
 import netaddr
 
-
-class SpiderFootTarget():
-    """SpiderFoot target.
-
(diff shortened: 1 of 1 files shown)


bcoles:
There's a lot happening in this commit and it's not clear what is being changed or why.

The code comments are intentional and should remain.

It appears the `matches` function has been removed entirely for some reason. This will break things.

Any changes to the classes also require changes to the associated unit and integration tests.

PR #1775, opened 2023-04-12
Improve README

AnonymousWP:
# Added

- Add steps to containerise SpiderFoot
- Add numbered list for steps

# Changed

- Remove blank lines in Dockerfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,28 +94,51 @@ To install and run SpiderFoot, you need at least Python 3.7 and a number of Pyth
 
 #### Stable build (packaged release)
 
-```bash
- wget https://github.com/smicallef/spiderfoot/archive/v4.0.tar.gz
- tar zxvf v4.0.tar.gz
- cd spiderfoot-4.0
- pip3 install -r requirements.txt
- python3 ./sf.py -l 127.0.0.1:5001
-```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -92,24 +92,75 @@ SpiderFoot's 200+ modules feed each other in a publisher/subscriber model to ens
...
-#### Stable build (packaged release):
+#### Stable build (packaged release)
+
+1. Download the package and unpack it
+
+    ```bash
+    wget https://github.com/smicallef/spiderfoot/archive/v4.0.tar.gz && tar zxvf v4.0.tar.gz && cd spiderfoot-4.0
+    ```
+
+1. Install the requirements
```
smicallef, line 103: Step numbering is not increasing here and below.

AnonymousWP: Are you sure you're not confused? As you can see in my branch, it does: https://github.com/AnonymousWP/spiderfoot/tree/Improve-readme#stable-build-packaged-release. Maybe you think it should increase the number, but some actions (e.g. "Docker stable build (packaged release)" for example) have only one step to be followed. Because of the small header, it's unclear it's about a new step maybe. That's why I also recommend starting with H1 -> H2 -> H3, etc.

PR #1779, opened 2023-04-19
sfp_deepinfo: Add Deepinfo module

ceylanb:
Deepinfo module can be used to find subdomains of a target domain by searching through the Deepinfo databases.


namtacs:
InternetDB (#1824) does not require authentication, provides all the data that this API provides, and also gives open ports and CVEs.
So, this is basically worse.

ceylanb:
@namtacs I couldn't understand your point. This PR is not related to InternetDB and there is nothing related to InternetDB in it. Could you please give more details for your point?

Issue #1820, opened 2023-11-04
sfcli: -e argument: AttributeError: 'str' object has no attribute 'close'


bcoles:
The `-e` argument clobbers `stdin` with file contents as a string:

https://github.com/smicallef/spiderfoot/blob/e84db07ffa681f1acd9445a2ffdc9f28ffe0330a/sfcli.py#L1368-L1379

Where as Python `Cmd` expects `_io.TextIOWrapper` not `str`:

```
# ./sfcli.py -e /tmp/asdf
Traceback (most recent call last):
  File "/root/Desktop/spiderfoot/./sfcli.py", line 1422, in <module>
    s.cmdloop()
  File "/usr/lib/python3.11/cmd.py", line 132, in cmdloop
    line = self.stdin.readline()
           ^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'readline'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/Desktop/spiderfoot/./sfcli.py", line 1424, in <module>
    cin.close()
    ^^^^^^^^^
AttributeError: 'str' object has no attribute 'close'
```

KDreynolds:
PR #1826  has been reviewed and is ready to merge to address this issue.

PR #1826, opened 2023-12-05
fix for issue #1820

KDreynolds:
quick fix for Issue #1820 

block of code reads the file specified by the -e argument into a string, which is then passed to the SpiderFootCli class as the stdin argument. However, the SpiderFootCli class expects an object with a readline method, not a string.

To fix this issue, I have modified the code to pass the file object itself, rather than a string containing the file's contents. Here's a rough idea of what the corrected code might look like:

### DIFF
--- a/sfcli.py
+++ b/sfcli.py
@@ -1368,8 +1368,7 @@ if __name__ == "__main__":
     # Load commands from a file
     if args.e:
         try:
-            with open(args.e, 'r') as f:
-                cin = f.read()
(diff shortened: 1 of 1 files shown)


KDreynolds:
Please let me know if I missed anything on this fix. Thanks!

howsmsmsuser (approved):

KDreynolds:
I believe this is ready to go, let me know if I need to do anything else here.

PR #1973, opened 2026-03-30
fix: ignore whitespace-only Disallow paths in extractUrlsFromRobotsTxt

### LINKED ISSUE #701: TODO: sflib.py: fix whitespace parsing; ie, " " is not a valid disallowed path
```python
        # We don't check the User-Agent rule yet.. probably should at some stage

        for line in robotsTxtData.splitlines():
            if line.lower().startswith('disallow:'):
                # todo: fix whitespace parsing; ie, " " is not a valid disallowed path
                m = re.match(r'disallow:\s*(.[^ #]*)', line, re.IGNORECASE)
                if m:
                    self.debug('robots.txt parsing found disallow: ' + m.group(1))
                    returnArr.append(m.group(1))
```

juliosuas:
fixes #701 (also the TODO in the docstring).

the regex is `r'disallow:\s*(.[^ #]*)'`. the leading `.` matches anything — including a space — so a bare `Disallow: ` ends up returning `[' ']` and a single-space "path" lands in the exclusion list.

changing the first char of the group to `\S` makes it require a non-whitespace start:

    r'disallow:\s*(\S[^ #]*)'

after:

    'Disallow: '              → []
    'Disallow: /admin'        → ['/admin']
    'Disallow: /p#comment'    → ['/p']

the spec treats empty `Disallow` as "allow everything" so returning `[]` there is the correct behaviour.

### DIFF
--- a/spiderfoot/helpers.py
+++ b/spiderfoot/helpers.py
@@ -931,8 +931,6 @@ class SpiderFootHelpers():
 
         Todo:
             Check and parse User-Agent.
-
-            Fix whitespace parsing; ie, " " is not a valid disallowed path
         """
         returnArr: typing.List[str] = list()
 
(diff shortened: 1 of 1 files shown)


juliosuas:
A bit more detail: per the [robots.txt spec](https://www.robotstxt.org/robotstxt.html), `Disallow:` with no path (or only whitespace) means "allow all" — it should not add any entry to the disallowed list. The old regex `r'disallow:\s*(.[^ #]*)'` used `.` which matches any character including space, so `Disallow: ` (space-only) was returned as `' '` — an invalid path. The `\S` fix ensures only paths starting with a real non-whitespace character are captured. Also removed the TODO comment from the docstring since it's now addressed.

juliosuas:
Closing this stale small fix to keep my open-source queue focused. Happy to revisit fresh if maintainers want this robots.txt behavior changed.

PR #2005, opened 2026-06-14
Fix/bugsquash codeql hardening

geeknik:


geeknik:
Sorry, misfired. Please disregard this.

PR #2027, opened 2026-08-13
fix: escape HTML output in scan result views (stored XSS)

### LINKED ISSUE #2012: Stored XSS via correlation result titles (OSINT-collected scan data rendered unescaped)
### Summary
SpiderFoot collects OSINT data from many untrusted external sources during a scan (a crawled page's content, a server's version banner, WHOIS records, document metadata, hostnames of co-hosted/affiliated systems, and so on). When the correlation engine produces a result, it builds the correlation title by substituting these raw, attacker-influenceable event values into a rule headline template (for example "Cloud storage bucket found open: {data}"). That title is stored and later returned by the `/scancorrelations` JSON endpoint without any HTML escaping, and the scan view template injects it directly into the DOM with jQuery `.append()`.

An attacker who controls a resource that SpiderFoot scans (their own server banner, WHOIS record for a domain that appears as an affiliate, a document served by the target, an open bucket name, etc.) can place an HTML/JavaScript payload in a field that ends up in a correlation title. When the operator opens the Correlations tab for that scan, the payload executes in the operator's browser. SpiderFoot ships with authentication disabled by default, and the operator's session can read the Settings page, which stores plaintext API keys for many third-party services. The XSS therefore allows exfiltration of those stored API keys and full control of the authenticated UI (start/stop/delete scans, change settings).

This is a sibling-sink to an existing server-side fix: the scan data element endpoint `/scaneventresults` does call `html.escape()` on the same event values (sfwebui.py line 1775), but the correlation endpoint and its render path were missed.

### Details

1) Correlation title is built from raw event data with no escaping.

`spiderfoot/correlation.py`, `build_correlation_title()` (lines 897-927):

```python
title = rule['headline']
if isinstance(title, dict):
    title = title['text']

fields = re.findall(r"{([a-z\.]+)}", title)
for m in fields:
    try:
        v = self.event_extract(data[0], m)[0]
    except Exception:
        self.log.error(f"Field requested was not available: {m}")
    title = title.replace("{" + m + "}", v.replace("\r", "").split("\n")[0])
return title
```

`v` is a raw scan-event value (OSINT-collected). It is substituted into the title with no HTML escaping. `create_correlation()` (lines 929-956) then stores this title via `correlationResultCreate()`.

38 of the shipped correlation rules embed raw event data into the headline, e.g.:
- `correlations/cloud_bucket_open.yaml`: `headline: "Cloud storage bucket found open: {data}"`
- `correlations/open_port_version.yaml`: `headline: "Software version revealed on open port: {data}"` (server version banner, attacker-controlled)
- `correlations/data_from_docmeta.yaml`: `headline: "Interesting data was found within document meta data: '{child.data}'"`
- `correlations/human_name_in_whois.yaml` / `email_in_whois.yaml`: WHOIS fields
- `correlations/http_errors.yaml`: `headline: "Multiple failure HTTP codes found at {entity.data}"`

2) The server returns the title unescaped.

`sfwebui.py`, `scancorrelations()` (lines 1722-1743):

```python
for row in corrdata:
    retdata.append([row[0], row[1], row[2], row[3], row[4], row[5], row[6], row[7]])
return retdata
```

`row[1]` is the correlation title. No `html.escape()` is applied to any column. Contrast the sibling endpoint `scaneventresults()` (lines 1771-1785), which DOES escape the same OSINT values:

```python
retdata.append([
    lastseen,
    html.escape(row[1]),   # <- data element value escaped here
    html.escape(row[2]),
    ...
])
```

3) The template injects the unescaped title into the DOM.

`spiderfoot/templates/scaninfo.tmpl`, `browseCorrelations()` (lines 459-492):

```javascript
sf.fetchData('${docroot}/scancorrelations', {'id': instanceId}, function(data) {
    ...
    for (var i = 0; i < data.length; i++) {
        table += "<tr>";
        table += "<td><a style='cursor: pointer' onClick='toggleCorrelation(\"" + instanceId + "\",\"" + data[i][0] + "\")'>" + data[i][1] + "</a>&nbsp;&nbsp;";   // line 471: data[i][1] = title, raw
        table += "<i class='glyphicon glyphicon-question-sign' ... data-title=\"" + data[i][5] + "\"></i>"; // line 472
        ...
    }
    ...
    $("#mainbody").append(table);   // line 492: DOM injection of raw HTML string
});
```

`data[i][1]` (the title) is concatenated straight into an HTML string and passed to jQuery `.append()`, which parses and inserts it as HTML. A title containing `` becomes a live element and the `onerror` handler runs.

### PoC

Steps (no special configuration required; default install, default no-auth):

1. Boot SpiderFoot: `python sf.py -l 127.0.0.1:5001`

2. An attacker arranges for a scanned resource to produce an event value containing an HTML payload. For example, with the `open_port_version` rule, the attacker runs a service whose banner is:

   `Apache/2.4`

   or for `cloud_bucket_open`, an open bucket whose name is:

   `evil-bucket`

   When SpiderFoot scans the attacker resource and the corresponding correlation rule fires, the correlation title becomes `... open: evil-bucket`.

3. The operator opens the scan and clicks the Correlations tab. The browser fetches `/scancorrelations`, which returns the title unescaped, and `scaninfo.tmpl` injects it via `.append()`. The `onerror` handler executes in the operator's session.

The full server path was confirmed end-to-end. Driving the real `build_correlation_title()`, the real DB layer, and the real running HTTP endpoint produced:

Server response from `GET /scancorrelations?id=<scan>`:
```json
[["937e8c39...","Cloud storage bucket found open: <img src=x onerror=alert(document.cookie)>evil-bucket","cloud_bucket_open","HIGH","Open cloud bucket","An open cloud storage bucket was found.","id: cloud_bucket_open\n...",1]]
```

The same event value via `GET /scaneventresults?id=<scan>` is escaped (`&lt;img src=x onerror=alert(document.cookie)&gt;evil-bucket`), confirming the correlation path is the missed sink.

Feeding the real server response through the exact `scaninfo.tmpl:471` concatenation yields the HTML string handed to jQuery `.append()`:
```html
<td><a ... onClick='toggleCorrelation("...","937e8c39...")'>Cloud storage bucket found open: <img src=x onerror=alert(document.cookie)>evil-bucket</a>...</td>
```
which contains a live `` element.

### Impact
Stored cross-site scripting. The attacker controls a resource that SpiderFoot collects data from (a server banner, an open bucket, a WHOIS record for a domain that shows up as an affiliate, document metadata served by the target, a hostname, etc.). The victim is the SpiderFoot operator who views the scan's correlations. Because SpiderFoot ships with authentication disabled by default and runs same-origin with the Settings page that stores plaintext third-party API keys, the executing JavaScript can read and exfiltrate those API keys and fully drive the authenticated UI (create, stop, delete scans, modify settings). The scope is changed (the injected data crosses from scan content into the operator's trusted UI origin), hence the elevated CVSS.

Affected Versions: 4.0.0 and earlier

TrueFurina:
### Summary

Stored XSS (fix #2012): OSINT-collected data is rendered into `innerHTML`
without escaping across **all** scan result views in `scaninfo.tmpl`:

- `browseEventData()` event table — `data[i][1]`/`data[i][2]` via
  `sf.replace_sfurltag()`, plus `data[i][0]`/`data[i][3]`/`data[i][8]`
- `viewScanLog()` — `data[i][0..3]` (log lines can contain collected data)
- `browseEventList()` — `data[i][1]` as link text
- `browseCorrelations()`/`toggleCorrelation()` — correlation titles,
  tooltips, event data
- unique-view (`browseEventData` unique mode) — `data[i][0]`/`data[i][2]`

`sf.replace_sfurltag()` only converts `<sfurl>` markers; it does not escape
HTML, so `<script>`/`` collected from a scanned target
execute in the SpiderFoot web UI context.

### Fix

- Add `sf.escapeHtml()` (escapes `& < > " '`) to
  `spiderfoot/static/js/spiderfoot.js`.
- Wrap every OSINT-data concatenation point in `scaninfo.tmpl` (23 calls)
  with `sf.escapeHtml(...)`; where `sfurl` links are intended, apply escape
  **first**, then `sf.replace_sfurltag(...)` so links still render (the
  escaped `&lt;sfurl&gt;` marker is handled by its existing first branch).
- Intentionally untouched (not OSINT-rendered): numeric id attributes
  (`id='cb_...'`, `corrwell_tr_...`) and the `escape()`-encoded value inside
  the `onClick` string in `browseEventList`.

### Verification

- `node --check spiderfoot/static/js/spiderfoot.js` → OK
- Behavior test running payloads through the exact rendering path:
  `<script>alert(1)</script>`, `">`,
  `';alert(1);//`, `<b>bold</b>`, `normal & "quotes"` → all escaped;
  `<sfurl>https://evil.example/p?a=1&b=2</sfurl>` → still renders as a link
- `grep` audit: no bare `data[i]` concatenation remains in `scaninfo.tmpl`
  (only numeric id/attribute usages)

### Files changed

- `spiderfoot/static/js/spiderfoot.js` (+8: `sf.escapeHtml`)
- `spiderfoot/templates/scaninfo.tmpl` (all OSINT-data render points escaped)

Fixes #2012

### DIFF
--- a/spiderfoot/static/js/spiderfoot.js
+++ b/spiderfoot/static/js/spiderfoot.js
@@ -52,6 +52,14 @@ document.addEventListener("DOMContentLoaded", () => {
 
 var sf = {};
 
+// Escape HTML metacharacters in OSINT-collected data before it is rendered
+// into innerHTML, preventing stored XSS via correlation/event results.
+sf.escapeHtml = function (data) {
+  return String(data).replace(/[&<>"']/g, function (c) {
+    return { "&": "&amp;", "<": "&lt;", ">": "&gt;", '"': "&quot;", "'": "&#39;" }[c];
+  });
+};
+
 sf.replace_sfurltag = function (data) {
   if (data.toLowerCase().indexOf("&lt;sfurl&gt;") >= 0) {
     data = data.replace(
--- a/spiderfoot/templates/scaninfo.tmpl
+++ b/spiderfoot/templates/scaninfo.tmpl
@@ -196,15 +196,15 @@
                                 for (var i = 0; i < data.length; i++) {
                                     table += "<tr>";
                                     if (typeName == null) {
-                                        table += "<td><pre class='table-border-bg-inherit'>" + data[i][8] + "</pre></td>";
+                                        table += "<td><pre class='table-border-bg-inherit'>" + sf.escapeHtml(data[i][8]) + "</pre></td>";
                                     }
                                     table += "<td><pre class='table-border-bg-inherit'>";
-                                    table += sf.replace_sfurltag(data[i][1]);
+                                    table += sf.replace_sfurltag(sf.escapeHtml(data[i][1]));
                                     // for debug
                                     table += "</pre></td><td><pre class='table-border-bg-inherit'>";
-                                    table += sf.replace_sfurltag(data[i][2]);
-                                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + data[i][3];
-                                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + data[i][0];
+                                    table += sf.replace_sfurltag(sf.escapeHtml(data[i][2]));
+                                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + sf.escapeHtml(data[i][3]);
+                                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + sf.escapeHtml(data[i][0]);
                                     table += "</pre></td>";
                                     table += "</tr>";
                                 }
@@ -433,11 +433,11 @@
                 for (var i = 0; i < data.length; i++) {
                     table += "<tr>";
                     table += "<td><pre class='table-border-bg-inherit'>";
-                    table += sf.replace_sfurltag(data[i][1]);
+                    table += sf.replace_sfurltag(sf.escapeHtml(data[i][1]));
                     table += "</pre></td><td><pre class='table-border-bg-inherit'>";
-                    table += sf.replace_sfurltag(data[i][2]);
-                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + data[i][3];
-                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + data[i][0];
+                    table += sf.replace_sfurltag(sf.escapeHtml(data[i][2]));
+                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + sf.escapeHtml(data[i][3]);
+                    table += "</pre></td><td><pre class='table-border-bg-inherit'>" + sf.escapeHtml(data[i][0]);
                     table += "</pre></td>";
                     table += "</tr>";
                 }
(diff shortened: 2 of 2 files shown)


TrueFurina:

ubaid605:

